PR #13, opened 2017-08-11
Decrement the number of running connections after a failed request

maxcelima:
After 20 failed connections the prismic api stops responding. The error handler was not decrementing the running connections counter and was not processing the next request in the query


arnaudlewis:
fixed in 1.2.0. thanks for your contribution 👍

PR #38, opened 2018-04-09
Bugfix: Prismic uses ambient type definition for "Document"

cjol:
Files like `ApiSearchResponse` currently produce values of `Document` type, but `Document` is not imported. That means that, instead of using Prismic's definition of Document, the default definition is used instead. As a result, clients using the type definitions get the wrong type (in my case, the type representing HTML Documents, provided by `react`).  I don't know if this ever worked correctly, but I don't see how it could do.

This pull request fixes this by simple importing Document on all relevant files.

### DIFF
--- a/d.ts/ApiSearchResponse.d.ts
+++ b/d.ts/ApiSearchResponse.d.ts
@@ -1,3 +1,4 @@
+import { Document } from "./documents";
 export default interface ApiSearchResponse {
     page: number;
     results_per_page: number;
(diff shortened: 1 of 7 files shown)


cjol:
The Travis build on node versions 4 and 5 is failing for unrelated reasons; namely that a dependency (punycode) now requires node v6.

```
error punycode@2.1.0: The engine "node" is incompatible with this module. Expected version ">=6".
```

srenault:
Thanks again. A related PR https://github.com/prismicio/prismic-javascript/pull/45 has been merged to fix that.

PR #39, opened 2018-04-09
Make Document a generic class to encourage type-safety of contents

cjol:
Previously, the returned data from a Prismic API call just had type `any`. Now, users have the option to provide an output type to methods like `.getByUID()`, so that the resulting document is correctly typed. This avoids the need for `any` or explicit type coercions on the client's part.

Before: 
Variable typed as `any`

After:
Variable typed correctly

N.B. The data itself still comes from a dynamic source, so there is no guarantee that the type will be correct at run-time. You can see in the second example that static type-safety is still helpful for avoiding the use of non-existent variables.


srenault:
Hey @cjol, 

A related PR https://github.com/prismicio/prismic-javascript/pull/45 has been merged. I think it should help you.

cjol:
Hi there,

Thanks for the update. My other PR is indeed no longer necessary now that `Document` is correctly exported. This PR does something different though - it makes `Document` a generic type, allowing strong typing of `template.data` (you can see the main contribution of this PR here: https://github.com/prismicio/prismic-javascript/pull/39/commits/a09c4211f81bb3603f28041d4cc070a3d2ce7197)

I think this PR should thus remain open, as even with #45 is currently still not possible to get type safety for `template.data`.

PR #41, opened 2018-04-16
Add token to previewSession query

### LINKED ISSUE #40: v1.5 doesn't allow preview of unpublished files / changes
When additional credentials are not supplied v1.5 doesn't allow preview of unpublished pages, and shows only the currently published version when previewing pages which have changes.

I think that this is a bug because the `token` supplied by the preview page should take precedence over any global credentials. When using the token supplied I should *always* be able to get the specified page in its current state.

In the previewSession method:

```typescript
previewSession(token: string, linkResolver: (doc: any) => string, defaultUrl: string, cb?: RequestCallback<string>): Promise<string> {
  return this.httpClient.request<PreviewResponse>(token).then((result) => {
    // ...
      return this.getByID(result.mainDocument).then((document) => {
        if (!document) {
          cb && cb(null, defaultUrl);
          return defaultUrl;
        } else {
          const url = linkResolver(document);
          cb && cb(null, url);
          return url;
        }
      });
    //...
  })
  //...
}
```

The initial `previewSession` request with the supplied token returns the correct document ID in `result.mainDocument`, but then this ID is fed into `getByID` which uses `queryFirst`/`query`. In the `query` method, the `ref` property in the options is set to `this.masterRef.ref`, which is the resolved document ID, but the query returns nothing if the page is not published, or the current-published state of the page for a published page, but not it's updates.

If this is a bug, a working solution is to add an `options` object to the `getByID` call with the `ref` property set to the `token`, eg:

```typescript
return this.getByID(result.mainDocument, {ref: token})
```

In the previous version of `previewSession`, this was the case:

```typescript
previewSession(token: string, ...) {
  api.requestHandler.request(token, function(err: Error, result: any, xhr: any) {
    // ...
    api.everything().query(Predicates.at("document.id", mainDocumentId)).ref(token).lang('*').submit()
    // ... 
  })
}
```

I will also raise a pull request referencing this issue.

Trolleymusic:
Calls `this.getByID` with preview session token in the options object. Fixes #40


Trolleymusic:
The failing checks on Travis seem to be unrelated to this pull request, as others are having them too

Trolleymusic:
Ah, ok. No worries, it wasn’t on purpose. I think Atom might have just done that 😆

PR #45, opened 2018-05-08
Add missing imports

olegcherr:
I noticed that the `Document` interface is not imported in the classes that use it. So in my TypeScript-project all the things related to this interface are broken :(

Please merge this PR asap  :+1:


srenault:
Hello,

I'm not sure about your issue but the declaration files are generated. We shouldn't edit by hand them.

olegcherr:
I didn't edit `.d.ts` files by hand.
I fixed only source `.ts` files and then recompiled the project.
`.d.ts` files re-generated automatically.

srenault:
Ok got it. You've right, imports are missing.

PR #56, opened 2018-08-28
Fix Ref interface

idlefingers:
`isMasterRef` was described as a string, but the API returns a boolean, which made following the docs to configure in-site preview a pain because `api.refs.find(ref => ref.isMasterRef)` tries to compare a string to a boolean 😢


arnaudlewis:
@idlefingers thanks for the report, It'll be available in the next release

Issue #57, opened 2018-09-03
Consider removing polyfills


nsivertsen:
Maybe, for the upcoming 2.0 release, one could consider removing the `fetch` and `Promise` polyfills from `prismic-javascript`. They account for almost half of the bundle size. Both `fetch` and `Promise` have pretty good browser support by now, and most apps already have some kind of polyfill strategy in place (e.g. [polyfill.io](https://polyfill.io/)), meaning users are forced to download polyfills twice in many cases.

PR #58, opened 2018-09-03
Remove fetch and Promise polyfills

nsivertsen:
See [#57](https://github.com/prismicio/prismic-javascript/issues/57).

I feel like there must be a more elegant way to patch `fetch` in node, but maybe this is fine?


nsivertsen:
Sorry, this was silly. Will reopen when I have this figured out.

PR #69, opened 2018-12-11
Fixes #27 remove type form package.json file to run build in angular …

### LINKED ISSUE #27: Provide better Typescript definition files
Hi Prismic folks :)

I'm trying to use the nodejs package with its bundled d.ts files.
However, types are not properly imported when using the module.
It comes from several facts:
- First, this comes from the way the module is declared, ie. using a namespace. `declare namespace Prismic {...}` & `export = Prismic;`. Only what is in the namespace is visible if i'm not mistaken.
- As a library, the barel file (index.d.ts) should re-export all public APIs. When fetching the api object with `Prismic.getApi(...)`, I cannot write the proper response type since the `PrismicApi` type is hidden in the namespace.

It could be better to change the namespace to a plain object while re-exporting all sub modules. This way, both the Prismic object and all APIs are visible.

Last but not the least, a little quote from the TS documentation :

> Exporting a namespace from your module is an example of adding too many layers of nesting. While namespaces sometime have their uses, they add an extra level of indirection when using modules. This can quickly become a pain point for users, and is usually unnecessary.

What do you guys think ?

lmeunierRakuten:
…6 production mode

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
     "url": "https://github.com/prismicio/prismic-javascript.git"
   },
   "main": "dist/prismic-javascript.min.js",
-  "types": "./d.ts/index.d.ts",
   "scripts": {
     "build": "node_modules/.bin/webpack-cli -p",
     "dev": "node_modules/.bin/webpack-cli -d --watch",


arnaudlewis:
Hi @lmeunierRakuten , the `types` property is essential to export the typescript definitions as you can see here: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html.
We should find another buy to make it work with the production build of angular 6.

krabbypattified:
Why/how does `types` break Angular 6 in production mode?

lmeunierRakuten:
Because your declaration is not correct.
In Issues #27 @nimaen explain that.

with your declaration it is impossible to do this type of import

`import { getApi, Predicates } from 'prismic-javascript';`
or 
`import ResolvedApi from 'prismic-javascript/d.ts/ResolvedApi';`

I try 
`
import * as Prismic from 'prismic-javascript';

Prismic.getApi
Prismic.Predicates 
`
 But not work to.

lmeunierRakuten:
To use your code I am forced to use the option "allowSyntheticDefaultImports", I did not want to use it because it involves modifying the default configuration of an angular build. It will be necessary for me to explain that prismic does not export ES2015 and suddenly we are forced to use this option.

I'm afraid that our future client does not understand and we are asking for another solution. That would be a shame.

nsivertsen:
I agree that it is somewhat unfortunate that `prismic-javascript` exports its API underneath one default export:

```js
export default {
  getApi,
  Predicates,
  // ...
}
```

... rather than as named exports:

```js
export {
  getApi,
  Predicates,
  // ...
}
```

This also breaks tree-shaking.

PR #73, opened 2018-12-20
added documentation for including Prismic through CDN

mccare:
If you include Primsic through CDN (script tag), the variable name is PrimsicJS and you can leave the require out. I've added a section to the readme in ES6 syntax.


hypervillain:
This was definitely needed. Thanks @mccare!

PR #76, opened 2019-01-24
Added linked_documents into the document type

james-w-arnold:
Querying the Prismic Api V2 returns a document model which includes a linked_documents field. This field is not included in the Document type definition, causing issues when writing unit tests with and defining fixtures based on the API Data.


srenault:
Hello, the `linked_documents` issue has been deprecated in favor of fetchLinks (https://prismic.io/docs/javascript/query-the-api/fetch-linked-document-fields). This property doesn't even work properly.

PR #90, opened 2019-07-12
Improve type definition

adam187:
It looks like types are not consistent with the api response. 

https://prismic.io/docs/javascript/query-the-api/the-response-object


lihbr:

PR #96, opened 2019-08-06
Using unmock for nock + fixtures

mikesol:
Hi!

I'm one of the authors of unmock, and we have just released a new version that allows for easy testing of API integrations using fixtures.  I see that prismic is currently defining their fixtures in `.js` files, and I think your repo would be a really good candidate for unmock.  Unmock simplifies the management of fixtures, reduces the size of testing code, and makes it easier to do property-based testing for APIs using tools like fast-check.

It'd be great if you could review the PR and the documentation on [unmock.io](https://www.unmock.io) and, if it seems like a good fit, I'd love to help transition the fixture part of your codebase to unmock!

### DIFF
--- a/__unmock__/prismic/index.yml
+++ b/__unmock__/prismic/index.yml
@@ -1,0 +1,10 @@
+servers:
+- https://example.prismic.io
+paths:
+  /api/v2: {
+    refs: [],
+    experiments: {},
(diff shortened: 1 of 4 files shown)


lihbr:
fixed in v6 🎉

PR #97, opened 2019-08-27
CSP friendly dist

henkerik:
@arnaudlewis @MarcMcIntosh  Up-to-date version of https://github.com/prismicio/prismic-javascript/pull/60.

The current distribution(https://github.com/prismicio/prismic-javascript/blob/master/dist/prismic-javascript.min.js) contains the line: 

```
try{
  n=n||Function("return this")()||(0,eval)("this")
} catch(t) {
  "object" == typeof window && (n=window)
}
```

I guess this is due to the use of webpack:

https://github.com/webpack/webpack/issues/6461

This line uses `eval` (either directly or through `Function("")`) and if your CSP policy doesn't allow eval it will fail and fallback to the catch clause. This will result in a violation event which will be send to the `report-uri` configured in your CSP policy. The code above will be executed on every page load, so on *every* page load, our app will send a violation report.

I updated the PR of @nsivertsen because the rollup build doesn't contain the problematic code.


src/request.ts
```diff
@@ -1,3 +1,12 @@
+// In the browser, node-fetch exports self.fetch:
+// https://github.com/bitinn/node-fetch/blob/master/browser.js
+import fetch, { RequestInit } from 'node-fetch';
+
+// Number of maximum simultaneous connections to the prismic server
+const MAX_CONNECTIONS: number = 20;
```
arnaudlewis: These lines are related to some old code, it's not used here.

src/index.ts
```diff
@@ -1,28 +1,25 @@
...
-  export const Api = PrismicApi;
-
-  export function client(url: string, options?: ApiOptions) {
-    return new DefaultClient(url, options);
-  }
-
-  export function getApi(url: string, options?: ApiOptions): Promise<ResolvedApi> {
-    return DefaultClient.getApi(url, options);
-  }
+export function client(url: string, options?: ApiOptions) {
```
arnaudlewis: You should take the opportunity to add the return type here

alexanderfletcher:
any update on this, the dist not being CSP friendly is a big issue

MarcMcIntosh:
Hi @henkerik, we fixed this issue in https://github.com/prismicio/prismic-javascript/pull/104
Thanks for following up :)

Best,

Marc

PR #105, opened 2019-10-31
Rollup build (Version 3.0.0)

### LINKED ISSUE #57: Consider removing polyfills

### LINKED ISSUE #59: Output tree-shakeable module

MarcMcIntosh:
### Continuation of https://github.com/prismicio/prismic-javascript/pull/60

# Background

As mentioned in #59, `prismic-javascript` currently only exports a UMD build that is not [tree-shakeable](https://developer.mozilla.org/en-US/docs/Glossary/Tree_shaking) by bundlers and therefore forces consumers of the library to include everything in their bundles, instead of only the functions they actually use.

In addition, as mentioned in #57, `prismic-javascript` currently forces down polyfills for `fetch` and `Promise`. By now, everyone know that JS is the most expensive [[1](https://medium.com/dev-channel/the-cost-of-javascript-84009f51e99e)], [[2](https://medium.com/@addyosmani/the-cost-of-javascript-in-2018-7d8950fbb5d4)] resource on most websites. Since browser support for [fetch](https://caniuse.com/#feat=fetch) (87.1%) and [Promise](https://caniuse.com/#feat=promises) (88.85%) is actually quite good, and most real-world projects already have some polyfilling strategy (such as [polyfill.io](https://polyfill.io/v2/docs/)) in place (meaning that in the worst case, users are forced to download the polyfills twice), it feels a little unnecessary to include them here.

I've therefore removed `promise-polyfill` and replaced `cross-fetch` with `node-fetch` (which is used internally by `cross-fetch` when run on Node), which uses `window.fetch` when bundled for the browser. Even before tree-shaking, this already reduces the bundle size of `prismic-javascript.min.js` from 31.9KB to 13.5KB.

# Description

Aside from the existing `dist/prismic-javascript.min.js` UMD build, this PR adds the following builds:

- `dist/prismic-javascript.js`: CJS build for Node.js
- `dist/prismic-javascript.mjs`: ESM build for Node.js
- `dist/prismic-javascript.browser.js`: UMD build for browsers
- `dist/prismic-javascript.browser.mjs`: ESM build for browsers

It uses the [`package.browser`](https://github.com/defunctzombie/package-browser-field-spec) field, which is understood by all major bundlers (Webpack, Rollup, Browserify), to provide the browser bundles.

Since [Webpack still doesn't support outputting ES modules](https://github.com/webpack/webpack/issues/2933), I was forced to switch to Rollup, but I think you'll find the change quite pleasant as builds are faster and the configuration file is simpler (especially when considering it now outputs five bundles rather than one).

Unfortunately, because polyfills are no longer shipped with the bundle, this is potentially a breaking change for some people (but Semver protects us from that), but one I hope you'll agree is favourable in the long-term and might consider including in the upcoming 2.0 release. 🤞

There are some things left to do (see below), but I thought I'd leave this here already for discussion.

Thank you!

# Breaking changes
Before this PR the kit is including a polyfill of `fetch` in the bundle. This can make some other modules (such as https://github.com/birkir/gatsby-source-prismic-graphql) who executes `fetch` in a NodeJS env without polyfilling themselves fetch, to fail. 
Such module would need to polyfill themselves `fetch` (using `node-fetch` seems the best way then so it'd not be added twice).

# Todo

- [x] rename `dist/prismic-javascript.esm.js` to `dist/prismic-javascript.mjs` for Node compatibility
- [x] adapt tests to new build outputs
- [x] add section on polyfills to README
- [x] add section on bundling for node and browsers to README

Fixes #57, fixes #59


MarcMcIntosh:
# TODO
This could be a major version change

MarcMcIntosh:
Clean dist directory

MarcMcIntosh:
@srenault  the original PR author has been in contact to clear up some of the details. I think it would be a good idea to merge his branch into an upstream branch for a major version update.

MarcMcIntosh:
TODO:

+ [] review documentation and ensure the instructions are accurate.
+ [] fix travis so that build passes.
+ [] check merge strategy with master.
+ [] update major version.

README.md
```diff
@@ -84,11 +85,35 @@ https://unpkg.com/prismic-javascript
...
++ prismic-javascript.browser.js: UMD build for browsers
++ prismic-javascript.browser.mjs: ESM build for browsers
+
+#### Polyfills
+To support legacy browsers include a fetch and promise polyfills.
+For node environments include a fetch polyfill.
+
+#### Bundling 
+Webpack should work as is.
+If using Rollup remember and targeting browsers remember to turn on the browser flag in the [resolver options](https://github.com/rollup/plugins/tree/master/packages/node-resolve#browser)
```
veloce: Small typo: the first remember should not be here.

MarcMcIntosh: Thanks, fixed it :)

rollup.config.js
```diff
@@ -46,6 +46,6 @@ export default [
       }),
       prod && terser()
     ],
-    external: ['fetch']
+    external: ['node-fetch']
```
hypervillain, line 49: How would this behave with people using other flavours of fetch?

veloce: This config tells rollup not to include `node-fetch` in the bundle. Since this is the bundle that is targeting browsers, `window.fetch` will be used internally by node-fetch. The only thing we need to take care of is the compatibility with the node version so it works on both sides. List of differences is here: https://github.com/node-fetch/node-fetch/blob/master/docs/v2-LIMITS.md

Since this plugin previously used `cross-fetch` which internally uses `node-fetch` we know we're good.
The only change I made from the preview implem in this PR is that I included node-fetch in the node bundle since I felt it's not convenient and should not be the responsibility of a node user to include a fetch polyfill. So now it fixes a change that would have been breaking on node side, keeping a lighter bundle on browser side. Hope all this makes sense ;)

veloce:
To support backward compatibility I had to do this in the index file:
https://github.com/prismicio/prismic-javascript/blob/a4382a65a193c41a72fd2350e45b97edf488900e/src/index.ts#L7-L29

This was needed because with the version 2.x, projects can actually import like:
```
import Prismic from 'prismic-javascript'
```
or
```
import * as Prismic from `prismic-javascript`
```

This is not ideal. Since there is already a breaking change and this PR is a new major version, we could fix this as well here and either export the default or the named but not both.
What do you think?

rollup.config.js
hypervillain: Makes perfect sense, thanks a lot for your explanation

PR #107, opened 2019-12-10
Expose into a function the logic to get the ref

srenault:
Expose into a function the way we resolve the ref we're going to use (master ref, preview ref or experiment ref).
This exposed function is useful for `apollo-link-prismic`.


srenault:
Will do it in another way!

PR #108, opened 2019-12-13
Fix type definition of alternate_languages

srenault:
- Fix https://github.com/prismicio/prismic-javascript/issues/102
- Remove dist file from the repository
- `npm audit fix`

### DIFF
--- a/src/documents.ts
+++ b/src/documents.ts
@@ -1,3 +1,10 @@
+export interface AlternateLanguage {
+  id: string;
+  uid: string;
+  type: string;
+  lang: string;
+}
+
 export interface Document {
   id: string;
   uid ?: string;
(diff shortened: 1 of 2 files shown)


src/documents.ts
```diff
@@ -1,3 +1,10 @@
+export interface AlternateLanguage {
+  id: string;
+  uid: string;
```
pasieronen: This should be `uid?: string`, I think (based on experimenting with the api)

pasieronen:
I'm facing the same issue, and would like to see this bugfix merged and released. The fix looks good (with one nit, commented above, about `uid` type).

The file `d.ts/documents.d.ts` is generated from `src/documents.ts` as part of `npm run build (I'm not sure why it's committed to git, though - but it should be in the tarball published to npm).

All the current tests are written in JavaScript, not TypeScript, so they can't catch problems that occur only in TypeScript type checking, not run-time.

src/documents.ts
hypervillain: Indeed, `uid` is optional in Prismic

PR #115, opened 2020-01-22
Feature: Adds boolean value in type PredicateValue.

luxcem:
The api is able to handle Predicates with `boolean` values (https://prismic.io/docs/rest-api/query-the-api/query-by-a-field#14_0-a-boolean-field). But the `PredicateValue` does not contain `boolean` and the `encode` method does not encode booleans.


Rutulpatel7077:
This PR will probably fix #117 Error: Unable to encode true of the type boolean

PR #119, opened 2020-02-27
Deprecate unreliable mainDocument

srenault:
```
const Prismic = require('prismic-javascript');

exports.prismicPreview = async (req, res) => {
  const { token, documentId } = req.query;
  const api = await Prismic.api('http://your_repository_name.prismic.io/api');
  const redirectUrl = await api.getPreviewResolver(token, documentId).resolve(linkResolver, '/');
  cookies.set(...) // set cookie
  res.redirect(302, redirectUrl);
}
```
# TODO
- [x] Update the description PR.
- [x] Add some tests.
- [ ] Test it in a real project.

Description

the `previewSession `function can sometimes fail due to an unreliable mainDocument (document id field) returned by httpClient.request(token) causing the preview-url to default to "/"

This update provides a new function `getPReviewSession(token, documentId)` that should eliminate the need for using the token to fetch the document-id.


niackouai:
Documentation to update

PR #123, opened 2020-03-19
Implement response timeout

### LINKED ISSUE #55: set timeout
Hi,
Sometimes, data query takes too long, how can i set a timeout value ?

veloce:
Closes #55


.gitignore
```diff
@@ -163,4 +163,7 @@ typings/
 # End of https://www.gitignore.io/api/node,macos,intellij,visualstudiocode
 
 ## documentation
-docs/
\ No newline at end of file
+docs/
+
+## dev build
+dist/prismic-javascript.js
```
MarcMcIntosh: This will be ignored by npm too, is that what we want to do?

Heyrows: I would say yes because we want to have an up to date version any time we use this kit, not some old kept version that does not work with new development

MarcMcIntosh:
Minor version update, and rebuild?

PR #124, opened 2020-03-20
Add support for preview resolver to client

srenault:
Related to https://github.com/prismicio/prismic-javascript/pull/119


src/client.ts
```diff
@@ -70,6 +72,13 @@ export class DefaultClient implements Client {
     return this.getApi().then(api => api.previewSession(token, linkResolver, defaultUrl, cb));
   }
 
+  getPreviewResolver(token: string, documentId?: string): PreviewResolver {
```
Heyrows: What the purpose of having this function here ?
We have the same in the ResolvedApi, so for which purpose would we need this one ?

src/ResolvedApi.ts
```diff
@@ -231,26 +224,7 @@ export default class ResolvedApi implements Client {
...
-          }
-        });
-      } else {
-        return Promise.resolve(defaultUrl);
-      }
-    }
-    
-   
-    return { token, documentId, resolve }
+    return createPreviewResolver(token, documentId, this.getByID.bind(this));
```
Heyrows: would it be possible to have something like:
getPreviewResolver( a, b, c) = createPreviewResolver( a, b )

Or would the this not be available in this situation ?

PR #135, opened 2020-06-29
Fix getPreviewResolver

srenault:
When using the `Client` interface, it seems we never overrides the ref with the preview token.
This causes some trouble mainly when previewing unpublished documents (not redirected to the right page).


hypervillain:
👍

PR #136, opened 2020-07-07
fix: do not call external when using previewSession method

mjeanroy:


arnaudlewis:
hey @mjeanroy, thanks for your contribution.
A question though, what does this PR fix? Have you had this case of an external URL?
The token generation is controlled so I don't get why we need a check here?
Regarding the code, the logic seems good but that would be nice to make some small changes to have a more readable code.
Using a regexp and naming the ouput variables would make it shorter and clearer here than using a bunch of slice/indexOf

mjeanroy:
Hi @arnaudlewis,
Thanks for the response.
I would prefer not talking too much the reason of this PR publicly (that's why I did not make any comment).
Is it possible to contact you by email?
You should also talk to your colleague Léonard West, I explained him by email few days ago.

arnaudlewis:
Hello @mjeanroy,
about this issue, the `previewSession` was simply deprecated for some time now and since we're about to release a new major version, we simply dropped it to keep only the `getPreviewResolver`.
In a near future, all of this will become unnecessary due to a new way to manage the link resolving that we put in place with SliceMachine already.
You can find the changelog here: https://github.com/prismicio/prismic-javascript/releases/tag/v4.0.0-beta.4

PR #138, opened 2020-07-17
Fix type reference in index.d.ts

### LINKED ISSUE #133: TypeScript build-time error: `Cannot find type definition file for 'types/cross-fetch'`
When attempting to build a TypeScript project that has `prismic-javascript@3.0.0` as a dependency, I get the following error:

```bash
node_modules/prismic-javascript/types/index.d.ts:1:23 - error TS2688: Cannot find type definition file for 'types/cross-fetch'.

1 /// <reference types="types/cross-fetch" />
                        ~~~~~~~~~~~~~~~~~


Found 1 error.
```

Is [the triple-slash directive in `index.d.ts`](https://github.com/prismicio/prismic-javascript/blob/3.0.0/types/index.d.ts#L1) supposed to be:

```ts
/// <reference types="cross-fetch" />
```
?

pvdstel:
Resolves #133.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,4 +1,4 @@
-/// <reference types="types/cross-fetch" />
+/// <reference types="./cross-fetch" />
 import ResolvedApi from './ResolvedApi';
 import { Experiments } from './experiments';
 import Api, { ApiOptions } from './Api';


pvdstel:
Any news on this?

veloce:
Fixed in a different way in b998175b. Since this is a generated file it couldn't be edited directly.

PR #143, opened 2020-10-06
V4

arnaudlewis:


lihbr:
Kit has been in production for a while now and is referenced on the doc, when are we merging? :)

PR #146, opened 2020-11-04
Improve types for Document and ApiSearchResponse

jsfelix:


lihbr:

PR #153, opened 2021-02-05
Use Authorization header to send access token

Raph22:
Instead of using `access_token` query parameter to pass the access token, we now use the Authorization header like that:
```
Authorization: Token MY_SUPER_TOKEN
```
fixes prismicio/issue-tracker-wroom#351

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -162,6 +162,9 @@ typings/
 
 # End of https://www.gitignore.io/api/node,macos,intellij,visualstudiocode
 
+### IntelliJ
+*.iml
+
 ## documentation
 docs/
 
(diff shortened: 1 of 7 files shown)


francois-riviere:
fixes https://github.com/prismicio/issue-tracker-wroom/issues/351

arnaudlewis (approved):
seems good on the code part but a test is necessary @francois-riviere

bapmrl (changes requested):
Please do not merge as is because the CORS preflight request fails. See this [comment](https://github.com/prismicio/issue-tracker-wroom/issues/351#issuecomment-812010428).

lihbr:
Related: https://github.com/prismicio/prismic-javascript/issues/168

lihbr:

PR #187, opened 2021-08-12
feat: add support httpOptions for fetch

dhoko:
Hey :wave: 

We use the gatsby-source-prismic plugin, it doesn't work inside our CI because there is no support  for Agent inside (a PR is coming). 

While I was reading the source I found this one doesn't support agent anymore (v5 does).

cf https://nodejs.org/api/https.html#https_class_https_agent we need this to be able to use the plugin as we have a proxy inside our runner.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -7,6 +7,7 @@ import { getCookie } from "./lib/getCookie";
 import {
 	FetchLike,
 	Form,
+	HttpOptionsLike,
 	HttpRequestLike,
 	Query,
 	Ref,
@@ -168,6 +169,8 @@ export type ClientConfig = {
 	 * The function used to make network requests to the Prismic REST API. In environments where a global `fetch` function does not exist, such as Node.js, this function must be provided.
 	 */
 	fetch?: FetchLike;
+
+	httpOptions?: httpOptionsLike;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


angeloashmore:
Hey @dhoko, thanks for the PR!

You're correct; the upcoming v6 version of `@prismicio/client` does not support an HTTP Agent like v5. Instead, it allows you to provide a custom `fetch` function which can perform any modification necessary, including using an Agent.

You can see an example of overriding the `fetch` function here: https://github.com/prismicio/prismic-javascript/blob/f618d626aa74272b7c5a4f4da51c808e30617741/examples/custom-caching/index.js

More specific to your case, this is the most basic example of a custom fetcher:

```
import fetch from "node-fetch";

const client = prismic.createClient(endpoint, {
	fetch: async (url, options) => {
		// You can adjust options here including adding an agent.
		// Modifying the URL can also happen here if necessary.
		return await fetch(url, options)
	},
});
```

`gatsby-source-prismic`, which uses `@prismicio/client` v6 internally, does not provide a way to override the fetch function. We will need to add a `fetch` plugin option that is passed to `@prismicio/client` (currently is defaults to `node-fetch` without any options).

I will respond to your PR there (https://github.com/prismicio/gatsby/pull/424).

I'm going to close this PR as the library already supports a way to use an HTTP Agent, but please feel free to continue the conversation if there is something we haven't considered.

PR #193, opened 2021-10-01
feat: throttle getAll* methods and rename getAll to dangerouslyGetAll

angeloashmore:
## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Description

This adds throttle-like behavior to all `getAll*` methods. `500ms` is used as the duration between network requests.

This ensures API performance is sustainable even with large queries. `getAll*` methods should be treated as long-running requests that may make multiple network requests in a safe timely manner.

The following algorithm is used:

1. Query for a page of results.
2. **If the result tells us there are additional pages**:
   1. Wait the throttle threshold duration (500ms).
   2. Return to Step 1 with the next page.
   
   **If there are no additional pages**:
   1. Return the aggregate list of pages.

For example, with a 3 page query:

```
Request page 1 - 250ms
[wait 500ms]
Request page 2 - 350ms
[wait 500ms]
Request page 3 - 250ms

DONE
Total time: 1850ms
```

With a 1 page query:

```
Request page 1 - 250ms

DONE
Total time: 250ms
```

This PR also renames `getAll` to `dangerouslyGetAll` to dissuade usage.

## Checklist:

🐢

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -84,6 +84,8 @@ const MAX_PAGE_SIZE = 100;
  */
 export const REPOSITORY_CACHE_TTL = 5000;
 
+const GET_ALL_THROTTLE_DELTA_MS = 1000;
+
 /**
  * A ref or a function that returns a ref. If a static ref is known, one can be
  * given. If the ref must be fetched on-demand, a function can be provided. This
@@ -563,16 +565,36 @@ export class Client {
 		const { limit = Infinity, ...actualParams } = params;
 		const resolvedParams = { pageSize: MAX_PAGE_SIZE, ...actualParams };
 
-		const result = await this.get<TDocument>(resolvedParams);
+		let page = 0;
+		let documents: TDocument[] = [];
+		let latestRequestTimestamp = Date.now();
 
-		let page = result.page;
-		let documents = result.results;
+		const recursiveGet = async () => {
+			const now = Date.now();
 
-		while (page < result.total_pages && documents.length < limit) {
-			page += 1;
-			const result = await this.get<TDocument>({ ...resolvedParams, page });
-			documents = [...documents, ...result.results];
-		}
+			if (
+				page === 0 ||
+				now - latestRequestTimestamp >= GET_ALL_THROTTLE_DELTA_MS
+			) {
+				page += 1;
+				latestRequestTimestamp = Date.now();
+				const result = await this.get<TDocument>({ ...resolvedParams, page });
+				documents = [...documents, ...result.results];
+
+				if (page < result.total_pages && documents.length < limit) {
+					await recursiveGet();
+				}
+			} else {
+				await new Promise((res) =>
+					setTimeout(async () => {
+						await recursiveGet();
+						res(undefined);
+					}, GET_ALL_THROTTLE_DELTA_MS - (now - latestRequestTimestamp)),
+				);
+			}
+		};
+
+		await recursiveGet();
 
 		return documents.slice(0, limit);
 	}
--- a/src/client.ts
+++ b/src/client.ts
@@ -5,13 +5,7 @@ import { appendPredicates } from "./lib/appendPredicates";
 import { castThunk } from "./lib/castThunk";
 import { getCookie } from "./lib/getCookie";
 
-import {
-	FetchLike,
-	HttpRequestLike,
-	// TODO: Uncomment when the Authorization header can be used
-	// @see Related issue - {@link https://github.com/prismicio/issue-tracker-wroom/issues/351}
-	// RequestInitLike
-} from "./types";
+import { FetchLike, HttpRequestLike, RequestInitLike } from "./types";
 import { buildQueryURL, BuildQueryURLArgs } from "./buildQueryURL";
 import { ForbiddenError, isForbiddenErrorAPIResponse } from "./ForbiddenError";
 import { ParsingError, isParsingErrorAPIResponse } from "./ParsingError";
@@ -517,10 +511,7 @@ export class Client {
 	): Promise<prismicT.Query<TDocument>> {
 		const url = await this.buildQueryURL(params);
 
-		// TODO: Uncomment when the Authorization header can be used
-		// @see Related issue - {@link https://github.com/prismicio/issue-tracker-wroom/issues/351}
-		// return await this.fetch<Query<TDocument>>(url, params);
-		return await this.fetch<prismicT.Query<TDocument>>(url);
+		return await this.fetch<prismicT.Query<TDocument>>(url, params);
 	}
 
 	/**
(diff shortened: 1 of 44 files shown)


src/client.ts
```diff
@@ -575,15 +585,30 @@ export class Client {
...
+
+		while (
+			(!latestResult || latestResult.page < latestResult.total_pages) &&
+			documents.length < limit
+		) {
+			const start = Date.now();
+			latestResult = await this.get<TDocument>({
+				...resolvedParams,
+				page: latestResult ? latestResult.page + 1 : undefined,
+			});
```
lihbr, line 599: Since we're now assuming this process is a long-running one, maybe it's worth getting and providing explicitly the master ref (if none other provided) so that it cannot change during the crawling process? (the 5000ms cache on it might get busted, if another document has been published master ref changes and pagination can be broken)

If so, does the same apply for Integration Field ref (not sure)?

lihbr (approved):
Awesome job! I just think there are some minor tweaks to be made but I really like the approach you took to throttle that neatly when in ideal conditions ☺️

src/client.ts
```diff
@@ -575,15 +585,30 @@ export class Client {
 		const { limit = Infinity, ...actualParams } = params;
 		const resolvedParams = { pageSize: MAX_PAGE_SIZE, ...actualParams };
 
-		const result = await this.get<TDocument>(resolvedParams);
+		const documents: TDocument[] = [];
```
srenault, line 588: I think we could simplify the code with the following function. The main idea is to use a generator:

```js
const browseAPI = async function*(url) {
  let nextUrl = url;

  while(nextUrl) {
    const response = await fetchPage(nextUrl);

    const { 'next_page': nextPage } = response;

    nextUrl = nextPage;

    yield response;
  }
};
```

The `getAll` function could rely on this and add a waiting time between each call.

src/client.ts
```diff
@@ -575,15 +585,30 @@ export class Client {
...
-			const result = await this.get<TDocument>({ ...resolvedParams, page });
-			documents = [...documents, ...result.results];
+			if (
+				latestResult.page < latestResult.total_pages &&
+				end - start < GET_ALL_THROTTLE_THRESHOLD
+			) {
+				await new Promise((res) =>
+					setTimeout(res, GET_ALL_THROTTLE_THRESHOLD - (end - start)),
+				);
+			}
```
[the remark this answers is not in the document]
srenault: I would say we can simplify with `setTimeout(res, GET_ALL_THROTTLE_THRESHOLD)`

src/client.ts
lihbr: Or maybe we can just use the `next_page` value from the previous query?

test/client-getAll.test.ts
```diff
@@ -124,3 +126,68 @@ test("merges params and default params if provided", async (t) => {
...
+	const endTime = Date.now();
+
+	const totalTime = endTime - startTime;
+	const maxTime = numPages * GET_ALL_THROTTLE_THRESHOLD;
+	const minTime = maxTime - GET_ALL_THROTTLE_THRESHOLD;
+
+	// The total time should be between (# of pages - 1) and # of pages
+	// multiplied by the throttle threshold duration. This effectively checks
+	// that each request after page 1 is delayed by at least the threshold
+	// amount.
```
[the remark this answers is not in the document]
angeloashmore: True, in the test I assume the request will be resolved quickly. Otherwise it may be difficult to test this without a deterministic delay.

MSW has a `delay` function that could make this more explicit: https://mswjs.io/docs/api/context/delay

src/client.ts
angeloashmore: Good idea, should be easy to grab the ref and reuse it for all requests. If someone publishes something during the fetching process, how long is a stale ref valid for?

`next_page` may be more work to use since ideally all queries can still go through the `this.get` method. `next_page` is a URL which would need to be parsed and processed.

src/client.ts
angeloashmore: Maybe I'm missing something, but I don't think a generator function adds anything here since can inline that logic into the method (i.e. no need for an abstraction). I'll update the code to use a static delay between requests which will result in simpler code.

test/client-getAll.test.ts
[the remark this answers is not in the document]
angeloashmore: The mock request handlers now use `delay` so we can more deterministically track how long a request will take to resolve.

We still need to account for the time it takes for the JS to run, so `NETWORK_REQUEST_DURATION_TOLERANCE` is defined at the top of the test file. Basically `NETWORK_REQUEST_DURATION_TOLERANCE` is the amount of time it takes for the method to complete its work between network requests, which will be machine-specific.

test/client-getAll.test.ts
```diff
@@ -124,3 +126,68 @@ test("merges params and default params if provided", async (t) => {
...
+
+	const startTime = Date.now();
+	await client.getAll();
+	const endTime = Date.now();
+
+	const totalTime = endTime - startTime;
+
+	// The total time should be less than throttle threshold duration. This
+	// effectively checks that the first request is called immediately and does
+	// not unnecessarily wait after the request is fulfilled.
```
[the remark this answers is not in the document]
angeloashmore: Updated the comment to reflect the simpler logic:

```
	// The total time should only be the amount of time it takes to resolve the
	// network request (accounting for some tolerance). In other words, there is
	// no artificial delay in a single page `getAll` query.
```

src/client.ts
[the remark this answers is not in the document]
angeloashmore: Done! There is now just a static delay between network requests.

I reduced the delay from 1000ms total (including network request) to 500ms (excluding the network request) between API requests. If a page takes 250ms to resolve, this is still 750ms between requests which is lengthy.

@srenault Is that enough time between API calls? Anything longer will be noticeable by users (for example, Gatsby users will be calling `getAll` every time they start the development server).

src/client.ts
angeloashmore: I didn't make this change yet because I want to make sure using a stale ref is okay. If a `getAll` request takes 10 seconds, is it better to use an updated ref half-way through the loop?

src/client.ts
angeloashmore: The full looping logic is now as follows:

```ts
const documents: TDocument[] = [];
let latestResult: prismicT.Query<TDocument> | undefined;

while (
	(!latestResult || latestResult.next_page) &&
	documents.length < limit
) {
	const page = latestResult ? latestResult.page + 1 : undefined;

	latestResult = await this.get<TDocument>({ ...resolvedParams, page });
	documents.push(...latestResult.results);

	if (latestResult.next_page) {
		await new Promise((res) => setTimeout(res, GET_ALL_QUERY_DELAY));
	}
}

return documents.slice(0, limit);
```

This retains use of `this.get` as a fetching and typing abstraction.

angeloashmore:
Thanks for the review @lihbr @srenault! I made some changes and commented on your comments.

src/client.ts
lihbr: I think refs are busted once they are N refs old (N new refs have been created since), we can confirm that tonight ☺️

OK indeed `next_page` is not so straightforward to use!

test/client-getAll.test.ts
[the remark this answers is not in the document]
lihbr: Great job! Yeah I was nitpicking sorry ^^'

angeloashmore:
This was approved by @srenault outside GitHub. 🎉

PR #195, opened 2021-10-04
fix: use MAX_PAGE_SIZE when pageSize param is nullish on getAll methods

angeloashmore:
## Types of changes

## Description

All `getAll*` methods default `pageSize` to `MAX_PAGE_SIZE` (currently 100). The default was only set if a `pageSize` parameter was not provided. For example, if a user provides a `pageSize` of `undefined`, `pageSize` will be `undefined`, not the default.

With this PR, `pageSize` will be set to `MAX_PAGE_SIZE` on any falsey value (including `0`, `undefined`, `null`, and no value).

## Checklist:

🦧

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -573,7 +573,10 @@ export class Client {
 		params: Partial<Omit<BuildQueryURLArgs, "page">> & GetAllParams = {},
 	): Promise<TDocument[]> {
 		const { limit = Infinity, ...actualParams } = params;
-		const resolvedParams = { pageSize: MAX_PAGE_SIZE, ...actualParams };
+		const resolvedParams = {
+			...actualParams,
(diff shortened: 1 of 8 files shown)


lihbr (approved):
I don't get why you moved the `pageSize` parameter below the `lang` one 🤔 is it only for styling purposes?

LGTM

angeloashmore:
> I don't get why you moved the pageSize parameter below the lang one 🤔 is it only for styling purposes?

It's due to the order of URL parameters. In the test, it compares the URL parameters as a string so order matters. It probably should convert it to an object and compare it (maybe `t.deepEqual` works on URLSearchParams).

Thanks for the review - I'll merge and publish.

lihbr:
Oh OK, makes sense, works this way, no worries!

PR #196, opened 2021-10-13
refactor: optimize code for bundle size

angeloashmore:
## Types of changes

(I guess a refactor is like a chore?)

## Description

Refactors the client to optimize bundle size. There likely are still areas that can be optimized further, but this is a good first effort and brings the code below 3 KB.

- **Before PR**: 3250 B (minified, gzipped, excluding `@prismicio/helpers`)
- **After PR**: 2921 B (minified, gzipped, excluding `@prismicio/helpers`)

Note: This PR removes the ability to provide a custom Integration Fields ref. This is something that should not be user-facing so it was removed.

## Checklist:

🐇

### DIFF
--- a/src/buildQueryURL.ts
+++ b/src/buildQueryURL.ts
@@ -175,14 +175,9 @@ export const buildQueryURL = (
 	endpoint: string,
 	args: BuildQueryURLArgs,
 ): string => {
-	const { ref, integrationFieldsRef, predicates, ...params } = args;
+	const { predicates, ...params } = args;
 
 	const url = new URL(`documents/search`, `${endpoint}/`);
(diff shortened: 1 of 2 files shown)


angeloashmore:
I was able to reduce the bundle size further by refactoring how predicates were exported. We were already exporting them using `* as predicate` which converts it to a single object, so we can do it manually to save a lot of bytes.

This means we can move the inline `await`s and restore the RefStateType enum (renamed to RefStateMode) without going over 3 KB.

With the latest changes, we are down to 2921 KB. Good enough to stop optimizing for now I think!

angeloashmore:
@lihbr I'll let you take a last look at it before merging.

PR #208, opened 2022-01-14
fix: resolve previewed document URL any language

AriFreyr:
## Types of changes

## Description

Preview only works for the default language, fetch the preview document for all languages. Otherweise getByID will crash before calling the linkResolver

## Checklist:

- [ ] All new and existing tests are passing.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1167,6 +1167,7 @@ export class Client {
 		if (documentID != null) {
 			const document = await this.getByID(documentID, {
 				ref: previewToken,
+				lang: "*",
 			});
 
 			// We know we have a valid field to resolve since we are using prismicH.documentToLinkField
--- a/test/client-resolvePreviewUrl.test.ts
+++ b/test/client-resolvePreviewUrl.test.ts
@@ -31,6 +31,7 @@ test.serial("resolves a preview url in the browser", async (t) => {
 		createMockQueryHandler(t, [queryResponse], undefined, {
 			ref: previewToken,
 			q: `[[at(document.id, "${documentId}")]]`,
+			lang: "*",
 		}),
 	);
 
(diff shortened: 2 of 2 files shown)


angeloashmore:
Hi @AriFreyr, thanks for the PR!

This is a good catch. I wonder if this could cause problems in other parts of an app once it gets past `getByID()`. If you preview a document in language "B", for example, but the rest of app uses language "A", you may be mixing documents from other languages.

I can see how your change is necessary, even in that situation. If an app uses the URL to determine how the client is configured, for example, you would need to first get past `resolvePreviewURL()` for the document.

## Requested changes

Could you update the tests for `resolvePreviewURL()` in `test/client-resolvePreviewURL.test.ts`?

In each test, the following `lang: "*"` line should be added in the `createMockQueryHandler()` function:

```diff
  server.use(
    createMockRepositoryHandler(t),
    createMockQueryHandler(t, [queryResponse], undefined, {
      ref: previewToken,
      q: `[[at(document.id, "${documentId}")]]`,
+     lang: "*",
    }),
  );
```

I can also make these changes if you enable maintainer edits ([details here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests)).

Thanks!

AriFreyr:
> This is a good catch. I wonder if this could cause problems in other parts of an app once it gets past `getByID()`. If you preview a document in language "B", for example, but the rest of app uses language "A", you may be mixing documents from other languages.

Yeah i can see how it would be confusing, i think the only other way to address this (and maybe a better way, atleast for my use case) would be if the preview callback url provided the language as query param,

I fixed the tests, should be working now

angeloashmore:
Great! Thanks for updating the tests. I'll merge and publish this. 🙂

> I think the only other way to address this (and maybe a better way, atleast for my use case) would be if the preview callback url provided the language of the document as a query param.

I could see that being useful. Since the previewed document contains the language in the `lang` parameter, the language should be obtainable easily, albeit with a network request.

angeloashmore:
This has been published in `v6.1.0`.

Thanks again @AriFreyr!

PR #221, opened 2022-02-04
feat: support abortable requests via `AbortController`

### LINKED ISSUE #214: Support for AbortController/signals
It would be great if the library methods would be cancelable. It should be possible since the library is using `fetch`. 

The use case. The user enters SPA website with a slow internet which makes a direct request to Prismic API. The user changes his mind (before the requests were finished) and navigates to another page within SPA. The requests will still finish even though the user is on another page and there is no way at the moment to cancel them.

Example of canceled `fetch` with AbortController:
https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort

angeloashmore:
## Types of changes

## Description

This PR adds the ability to abort network requests made by the client using an [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController). It follows the standard `fetch()` method of providing an `AbortSignal` via the `signal` param.

Any client method that makes a network request can be cancelled by providing a `signal` property to its last parameter. This includes non-query methods like `getRepository()` and `buildQueryURL()`.

Aborting a client method looks like the following (assume a client has been created):

```ts
const controller = new AbortController();

setTimeout(() => controller.abort(), 100);

const result = await client.getByUID("page", "home", {
	signal: controller.signal,
});
```

The request will be aborted after 100ms if it is not resolved by then.

This is useful to set timeouts per-request (see the [Custom Timeout example](https://github.com/prismicio/prismic-client/tree/master/examples/custom-timeout) for a global timeout). It is also useful if an in-progress request is replaced by a newer one (for example, navigating to a different route).

Closes #214

## Checklist:

🙊

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,5 +1,7 @@
 /**
  * A universal API to make network requests. A subset of the `fetch()` API.
+ *
+ * {@link https://developer.mozilla.org/en-US/docs/Web/API/fetch}
  */
 export type FetchLike = (
 	input: string,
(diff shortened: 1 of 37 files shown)


src/types.ts
```diff
@@ -1,16 +1,42 @@
...
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	onabort: any;
+};
+
 /**
  * The minimum required properties from RequestInit.
  */
 export interface RequestInitLike {
 	headers?: Record<string, string>;
+	signal?: AbortSignalLike | null;
```
lihbr, line 39: Is it ever null?

lihbr (approved):
LGTM to me, feel free to proceed :)

src/types.ts
angeloashmore: I'm not sure if realistically it is ever `null`, but TypeScript's global type for `fetch()` includes `null`. Without it being added here, you get a type error.

https://github.com/microsoft/TypeScript/blob/3328feb7991f358e245088d48b64ad9da8f015e2/lib/lib.dom.d.ts#L1515-L1518

angeloashmore: I added a comment so it's clear why it's added. It does look pretty odd to include.

PR #227, opened 2022-03-10
fix: cache-bust GraphQL requests made with `graphQLFetch()` by including a `ref` URL parameter

angeloashmore:
## Types of changes

## Description

This PR cache-busts GraphQL requests made with `graphQLFetch()`. It does this by including the request's `ref` as a URL parameter.

The Prismic GraphQL API ignores the `ref` parameter; the API reads the ref from a header named `Prismic-ref`.

Without this PR, re-running a query may return a locally cached response, even when after the `ref` changes. This happens because caches typically do not take into account headers when caching requests.

Related: https://github.com/prismicio/apollo-link-prismic/issues/39

## Checklist:

🕸️

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1377,18 +1377,34 @@ export class Client {
 			}
 		}
 
-		// Compress the GraphQL query (if it exists) by removing
-		// whitespace. This is done to optimize the query size and avoid hitting the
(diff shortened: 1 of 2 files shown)


angeloashmore:
Published as an alpha at `v6.4.1-alpha.1`.

Issue #239, opened 2022-05-11
Accept Headers object when using `enableAutoPreviewsFromReq()`


angeloashmore:
### Is your feature request related to a problem? Please describe.

`enableAutoPreviewsFromReq()`, as the name suggests, enables automatic preview support from a server request. The request potentially contains the preview cookie which is used to detect an active preview session.

Some frameworks, like SvelteKit, use the native Web API's Headers interface. `@prismicio/client` is currently designed to work with Express-style server requests, which use a simpler object representation of Headers.

### Describe the solution you'd like

To better support frameworks using the Headers interface, `@prismicio/client` should accept a server request build upon Web APIs.

### Describe alternatives you've considered

`enableAutoPreviewsFromReq()` can be bypassed by using `queryContentFromRef()` directly. This requires fetching the preview ref manually in the server request's headers. While this only takes a few lines of code, it is inconvenient and is not declarative like `enableAutoPreviewsFromReq()`.

### Additional context

N/A

angeloashmore:
Implemented in #240.

PR #240, opened 2022-05-19
feat: accept Web API Request object in `enableAutoPreviewsFromReq()`

angeloashmore:
## Types of changes

## Description

This PR adds support for [Web API Request](http://developer.mozilla.org/en-US/docs/Web/API/Request) objects containing a [Web API Headers](http://developer.mozilla.org/en-US/docs/Web/API/Headers) instance. Frameworks like SvelteKit use this API rather than an Express-style request object which, unlike the Web API counterpart, contains plain-object-based data.

```typescript

const headers = new Headers();
headers.set("cookie", `io.prismic.preview=abc123`);

const request = {
  headers,
  url: 'https://example.com/preview?documentId=abc123&token=abc123'
};

const client = prismic.createClient("qwerty");
client.enableAutoPreviewsFromReq(request);

// The client will automatically read the `io.prismic.preview` cookie
// and URL parameters to enable preview sessions.
```

For more details, see #239.

## Checklist:

🦮

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1492,15 +1492,24 @@ export class Client {
 	 */
 	private async getResolvedRefString(params?: FetchParams): Promise<string> {
 		if (this.refState.autoPreviewsEnabled) {
-			let previewRef: string | undefined = undefined;
+			let previewRef: string | undefined;
+
+			let cookieJar: string | null | undefined;
 
 			if (globalThis.document?.cookie) {
-				previewRef = getCookie(cookie.preview, globalThis.document.cookie);
-			} else if (this.refState.httpRequest?.headers?.cookie) {
-				previewRef = getCookie(
-					cookie.preview,
-					this.refState.httpRequest.headers.cookie,
-				);
+				cookieJar = globalThis.document.cookie;
+			} else if (this.refState.httpRequest?.headers) {
+				if ("get" in this.refState.httpRequest?.headers) {
+					// Web API Headers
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -1492,15 +1500,24 @@ export class Client {
...
 			if (globalThis.document?.cookie) {
-				previewRef = getCookie(cookie.preview, globalThis.document.cookie);
-			} else if (this.refState.httpRequest?.headers?.cookie) {
-				previewRef = getCookie(
-					cookie.preview,
-					this.refState.httpRequest.headers.cookie,
-				);
+				cookieJar = globalThis.document.cookie;
+			} else if (this.refState.httpRequest?.headers) {
+				if ("get" in this.refState.httpRequest?.headers) {
```
lihbr, line 1510: Shouldn't we check with `instanceof Headers` or `typeof === "function"` instead? This could fail if there's a "get" header 🤔

lihbr (changes requested):
LGTM, just a small one to make it stronger ☺️

src/client.ts
angeloashmore: Yep, you're right. Will fix! 🙂

angeloashmore: I went with an extra `typeof === "function"` since `Headers` doesn't exist in Node. Duck-typing is simpler in this case.

lihbr: Awesome!

PR #249, opened 2022-07-15
refactor: optimize `getFirst()` and dependent methods

angeloashmore:
## Types of changes

## Description

This PR optimizes `getFirst()` by defaulting to only query one document. It does this by setting `pageSize=1`.

The following methods also benefit from this optimization since they use `getFirst()` within their implementation:

- `getByID()`
- `getByUID()`
- `getSingle()`
- `resolvePreviewURL()`

**Exceptions to this optimization**:

- If a `page` param is provided, `pageSize` is not set. `pageSize` affects the behavior of `page`, so we cannot automatically set `pageSize` to 1.
- If a `pageSize` param is provided, it takes priority.

## Checklist:

🐥

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -531,7 +531,12 @@ export class Client<
 	async getFirst<TDocument extends TDocuments>(
 		params?: Partial<BuildQueryURLArgs> & FetchParams,
 	): Promise<TDocument> {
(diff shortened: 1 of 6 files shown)


lihbr (approved):
LGTM 👏

PR #261, opened 2022-08-27
feat: optimize concurrent queries via shared network requests

### LINKED ISSUE #260: perf: dedupe `/api/v2` calls on concurrent search requests
### Is your feature request related to a problem? Please describe.

A best practice when querying multiple pieces of content is to run requests concurrently using `Promise.all`, potentially dividing the total query time by _N_.

```typescript
// Sequential, requests are performed one after the other
const home = await client.getByUID("page", "home");
const partials = await client.getSingle("partials");
const settings = await client.getSingle("settings");

// Concurrent, all requests are performed at the same time
const [home, partials, settings] = await Promise.all([
	  client.getByUID("page", "home"),
	  client.getSingle("partials"),
	  client.getSingle("settings")
]);
```

However, because Prismic API relies on a ref being provided by `/api/v2`, this causes the concurrent approach to query `/api/v2` _N_ times, instead of one (as the sequential approach is able to rely on the cached repository).

With that in mind, in the above example, the sequential approach, while being slower, performs only 4 network calls compared to 6 for the concurrent approach.

image

### Describe the solution you'd like

We could potentially store ongoing promises to `/api/v2` and have concurrent calls all rely on the same promise. Maybe this causes issues with abort controllers though 🤔

### Describe alternatives you've considered

**Temporary alternative**

A temporary alternative is to first get the cached repository in the user's code. However, because `getCachedRepository` is a private method, TypeScript will cry.

```typescript
await client.getCachedRepository();
const [home, partials, settings] = await Promise.all([
	  client.getByUID("page", "home"),
	  client.getSingle("partials"),
	  client.getSingle("settings")
]);
```

**Pattern alternative**

Prismic client could provide a "run multiple query helper, performing both optimizations (`getCachedRepository` and `Promise.all`) for the user, potentially also helping us advocate for the optimization.

```typescript
class Client {
	async concurrently<TQueries extends readonly unknown[] | []>(
		queries: TQueries
	): Promise<{ -readonly [K in keyof TQueries]: Awaited<TQueries[K]> }> {
		await client.getCachedRepository();
		return Promise.all(queries);
	}
	// ...
}

// In userland
const [home, partials, settings] = await client.concurrently([
	  client.getByUID("page", "home"),
	  client.getSingle("partials"),
	  client.getSingle("settings")
]);
```

**Ignore alternative**

Maybe this issue is something we can just ignore 🤔

**GraphQL alternative**

This is a use case where GraphQL could make sense as you could potentially query everything in 2 queries (one at `/api/v2` to get the ref, then one at `/graphql`), if only you didn't have to specify every single field each time...

### Additional context

`/api/v2` is really fast, and when performing short burst calls (I guess under 5-seconds-interval calls), it is actually cached on Cloudfront (according to headers).

Here's a test file showcasing the 3 approaches (sequential, concurrent, concurrent deduped), check the console & network tabs.
[clientPerformanceTest.zip](https://github.com/prismicio/prismic-client/files/9401731/clientPerformanceTest.zip)

angeloashmore:
## Types of changes

## To do before merging

- [x] Add tests for all methods using `Client.prototype.fetch()`
- [ ] Refactor code to reduce size increase

## Description

This PR optimizes concurrent queries by sharing network requests. If two or more concurrent network requests look identical (same URL and, if given, signal), only one network request is made.

The primary purpose of this optimization is to share calls to `/api/v2` (i.e. `getCachedRepository()`), which all queries perform before calling the `/api/v2/documents/search` endpoint. The implementation, however, allows any concurrent equivalent network requests to be shared.

The following contrived example demonstrates identical concurrent queries:

```typescript

const client = prismic.createClient("qwerty");

// Two network requests are made:
//
// 1x request to `/api/v2`
// 1x request to `/api/v2/documents/search`
await Promise.all([
  client.get(),
  client.get(),
  client.get(),
  client.get(),
  client.get(),
]);
```

The following more realistic example demonstrates practical concurrent queries:

```typescript
// Four network requests are made:
//
// 1x request to `/api/v2`
// 3x requests to `/api/v2/documents/search`
await Promise.all([
  client.getByUID("page", "about"),
  client.getSingle("navigation"),
  client.getSingle("settings"),
]);
```

Note how only one call to `/api/v2` is made. Without this PR, multiple concurrent calls to `/api/v2` are necessary since each query starts immediately without a cached `/api/v2` response.

```typescript
// Without this PR, six network requests are made:
//
// 3x request to `/api/v2`
// 3x requests to `/api/v2/documents/search`
await Promise.all([
  client.getByUID("page", "about"),
  client.getSingle("navigation"),
  client.getSingle("settings"),
]);
```

### Abortable queries via `AbortSignal`

All query methods are abortable (i.e. cancelable) via the `signal` parameter. An `AbortSignal`, usually controlled by an `AbortController`, can be aborted, causing the network request to be stopped and throw an error.

A query with a signal and an equivalent query without a signal _are not equivalent_. If the `AbortSignal` is cancelled, we don't want the signal-less query to abort as well. Thus, network requests cannot be shared.

```typescript
const controller = new AbortController();

// Four network requests are made:
//
// 2x request to `/api/v2`
// 2x requests to `/api/v2/documents/search`
await Promise.all([
  // Shared
  client.get(),
  client.get(),

  // Not shared
  client.get({ signal: controller.signal }),
]);
```

### Things to consider

This approach to implementing #260 provides optimizations that are general. All network requests, not just calls to `getCachedRepository()`, are optimized.

Using a general implementation may not be necessary or useful in real-world scenarios. Calling the same query within a single `Promise.all()` should almost never happen.

It is possible for a single client to be shared across multiple contexts which makes identical queries in quick succession, such as a static site generator that calls `client.getSingle("settings")` on each page.

The general implementation in this PR takes up more bytes than something specific to `getCachedRepository()`. We need to decide if the bytes are worth optimizing every, albeit infrequent, edge case.

Closes #260

## Checklist:


src/client.ts
```diff
@@ -1642,51 +1659,64 @@ export class Client<
...
-		// const accessToken = (params && params.accessToken) || this.accessToken;
-		// const options = accessToken
-		// 	? { headers: { Authorization: `Token ${accessToken}` } }
-		// 	: {};
+		let job: Promise<FetchJobResult>;
 
-		const res = await this.fetchFn(url, {
-			signal: params.signal,
-		});
+		if (this.fetchJobs[url] && this.fetchJobs[url].has(params.signal)) {
```
lihbr, line 1666: Maybe it's worth adding some comments here explaining why we're doing that and/or a link to this PR

src/client.ts
```diff
@@ -1642,51 +1659,64 @@ export class Client<
...
+			// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+			job = this.fetchJobs[url].get(params.signal)!;
+		} else {
+			this.fetchJobs[url] = this.fetchJobs[url] || new Map();
```
lihbr, line 1670: Maybe refactor to:

```typescript
this.fetchJobs[url] ||= new Map();
```

lihbr (approved):
LGTM

Maybe we'll get issues about disabling that behavior for some edge cases but I can't think of any so maybe we can wait for those before adding a flag :thinking:

I think the size increase is fine, not sure we can do much better if only focusing on the cached repo.

src/client.ts
lihbr: Actually I saw the 2nd commit my bad. Isn't Babel supposed to do its job here?

lihbr:
Do we want to still work on that or no? I feel like it'd still be beneficial

angeloashmore:
Yes, let's try to get this into v7. Sorry, it's my fault that this didn't proceed; I totally forgot about it! I'll finish this PR before we merge v7.

lihbr:
No pressure, but any update on this @angeloashmore?

src/client.ts
angeloashmore: Done!

```
		// `fetchJobs` is keyed twice: first by the URL and again by is
		// signal, if one exists.
		//
		// Using two keys allows us to reuse fetch requests for
		// equivalent URLs, but eject when we detect unique signals.
```

src/client.ts
angeloashmore: It's been a while since I made this commit, but I believe this is due to running without a transpiler (i.e. running on Node).

To keep backwards compatibility without much overhead, I think we should keep `||` rather than use `||=`, at least until Node.js >=15 is adopted.

angeloashmore:
All tests have been updated with a new `testConcurrentMethod()` test helper.

I messed up and made all the changes as part of the `v7` merge commit. :<

The implementation has *not* been refactored to reduce the size increase, but I think we can leave it as is for now. As we get closer to a v7 release with its full feature set, we can refactor and optimize for size.

lihbr (approved):
LGTM

src/client.ts
```diff
@@ -1642,51 +1659,64 @@ export class Client<
...
-					`Prismic repository not found. Check that "${this.endpoint}" is pointing to the correct repository.`,
-					url,
-					undefined,
-				);
-			} else {
-				throw new PrismicError(undefined, url, undefined);
-			}
+			job = this.fetchFn(url, {
+				signal: params.signal,
+			}).then(async (res) => {
```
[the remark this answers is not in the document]
angeloashmore: Fixed in this commit: https://github.com/prismicio/prismic-client/commit/92dd2e73d33b1ceb5fdcfc1faafea4fd3c7de8b6

We need to use `then` to process successful requests, but we can use `finally` to clear the fetch job.

PR #275, opened 2023-01-19
refactor: organize types, errors, and imports

angeloashmore:
## Types of changes

## Description

This PR organizes types, errors, and imports.

- **Types**: Co-locate types where they are used. For example, `LinkResolverFunction` was moved into `asLink()`.
- **Errors**: Move errors into a dedicated `src/errors` directory.
- **Imports**: Sort imports using `@trivago/prettier-plugin-sort-imports`. See [`createClient.ts`](https://github.com/prismicio/prismic-client/blob/70958c2c48be53638504e02ca69c78ce71a232ed/src/createClient.ts) for an example.

Imports were sorted and grouped using the following categories:

- Node built-ins
- Absolute imports (i.e. packages)
- `__testutils__` (for tests)
- `__fixtures__` (for tests)
- `lib`
- `types`
- `errors`
- Relative imports sorted from most distant to closest
- Siblings (i.e. `./`)
- `src` (for tests)

This list is generic enough to be reused across all Prismic packages if approved.

`@trivago/prettier-plugin-sort-imports`, a Prettier plugin, was selected over `eslint-plugin-simple-import-sort`, an ESLint plugin, due to its more configurable output.

## Checklist:

🐰


package.json
```diff
@@ -28,10 +28,10 @@
...
 		"./package.json": "./package.json"
 	},
-	"main": "dist/index.cjs",
-	"module": "dist/index.js",
-	"react-native": "dist/index.js",
-	"types": "dist/index.d.ts",
+	"main": "./dist/index.cjs",
+	"module": "./dist/index.js",
+	"react-native": "./dist/index.js",
+	"types": "./dist/index.d.ts",
```
lihbr, line 34: Why do we need `./` prefix now?

test/__setup__.ts
```diff
@@ -1,7 +1,7 @@
-import { afterAll, beforeAll, beforeEach, vi } from "vitest";
-import { createMockFactory, MockFactory } from "@prismicio/mock";
-import { setupServer, SetupServerApi } from "msw/node";
+import { MockFactory, createMockFactory } from "@prismicio/mock";
 import AbortController from "abort-controller";
+import { SetupServerApi, setupServer } from "msw/node";
+import { afterAll, beforeAll, beforeEach, vi } from "vitest";
```
lihbr, line 4: I don't know how you feel about it, but I tend to prefer test libraries to be listed at the very top, wdyt?

lihbr (approved):
LGTM, just a few minor ones above.

So we also want to replicate this setup on the template and other packages mid term right?

package.json
angeloashmore: The updated [`size-limit.cjs`](https://github.com/prismicio/prismic-client/blob/9fb5ae29ad7d0135b83389f254a17f7a3d4251f6/.size-limit.cjs) file reads `main`, `module`, and all `exports` (and perhaps it should read `react-native` as well). It dedupes the file list using a simple `Set`, which meant `dist/index.js` and `./dist/index.js` were treated as separate files.

I suppose we could add some `path.resolve()` logic

angeloashmore: The updated [`size-limit.cjs`](https://github.com/prismicio/prismic-client/blob/9fb5ae29ad7d0135b83389f254a17f7a3d4251f6/.size-limit.cjs) file reads `main`, `module`, and all `exports` (and perhaps it should read `react-native` as well). It dedupes the file list using a simple `Set`, which meant `dist/index.js` and `./dist/index.js` were treated as separate files.

I suppose we could add some `path.resolve()` logic to `size-limit.cjs` to make it smarter, but this solution was simpler.

lihbr: Makes sense to me! It's also more standard apparently so let's go

test/__setup__.ts
angeloashmore: Maybe it could be that any `devDependency` is its own group at the very top? We would need to change the Prettier config to a JS file so we can programmatically generate that list.

I don't have a strong opinion on sorting test deps to the top. If we go the `devDependencies` route, `node-fetch` might be added to the top if it's only used in tests, but not if it is used in the library.

lihbr: Hmm, I'd prefer the boring declarative way over some programmatic way, but I can see how inconsistencies can get frustrating. I don't know ;_; `/roll 2`
image

angeloashmore: Declarative sorting works for me. If we want to hoist `vitest`, `msw`, memfs`, etc. to the top, we can add this line to the config:

```js
["^(vitest|msw|memfs)(/|$)"]
```

lihbr: OK, I'm for doing nothing

Issue #274, opened 2023-01-03
Run CI with various TypeScript verions


lihbr:
### Is your feature request related to a problem? Please describe.

In 2022 we encountered a few nasty edge-case-y-TypeScript-issues related to some specific TypeScript versions (partly because TypeScript is complex and doesn't strictly follow SemVer)

### Describe the solution you'd like

To attempt to catch those issues early we could update the CI to run type tests against various TypeScript versions. [`zod`](https://github.com/colinhacks/zod) does so following a matrix strategy for example: https://github.com/colinhacks/zod/blob/master/.github/workflows/test.yml#L11-L27

Since this package is the only one with critical types, I don't think it's worth the overhead to port this solution to other SDKs or the template for now (or maybe just React/Vue)

### Describe alternatives you've considered

Doing nothing is a also valid option 🤷‍♀️

### Additional context

In an attempt to mitigate these issues we already started to systematically type check our projects as part of our test suite, e.g. https://github.com/prismicio/prismic-client/blob/v7/package.json#L49

PR #278, opened 2023-01-23
ci: run type tests against various TypeScript version

lihbr:
## Types of changes

## Description

This PR runs type tests across various version of TypeScript (4.5 through 4.9).

It fixes the few errors pointed out by those new tests.
- A minor constructor error
- Exceptionally uses `@ts-ignore` instead of `@ts-expect-error` for a specific TypeScript 4.9 issue (not sure about it, but apparently https://github.com/microsoft/TypeScript/issues/51501 got fixed on the upcoming 5.0 version, someone suggested to backport it to 4.9 but for now it has been ignored)

This PR also refactors a bit the CI architecture to accommodate those new tests. It's a bit more complex but about as fast when not running `size-limit` and faster when running it.

image

Resolves: #274

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,8 +5,6 @@ on:
     branches:
       - master
   pull_request:
-    branches:
-      - master
 
 jobs:
   ci:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,10 +3,78 @@ name: ci
 on: [push, pull_request]
 
 jobs:
-  ci:
+  prepare:
+    name: Prepare ({{ matrix.os}}, Node ${{ matrix.node }})
     runs-on: ${{ matrix.os }}
-
     strategy:
+      matrix:
+        os: [ubuntu-latest]
+        node: [14]
+
+    steps:
+      - name: Set up Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node }}
+
+      - name: Checkout
+        uses: actions/checkout@master
+
+      - name: Cache node_modules
+        id: cache
+        uses: actions/cache@v3
+        with:
+          path: node_modules
+          key: ${{ matrix.os }}-node-v${{ matrix.node }}-deps-${{ hashFiles(format('{0}{1}', github.workspace, '/package-lock.json')) }}
+
+      - name: Install dependencies
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: npm ci
+
+  suite:
+    needs: prepare
+    name: Suite ({{ matrix.os}}, Node ${{ matrix.node }})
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -1,17 +1,11 @@
 name: ci
 
-on:
-  push:
-    branches:
-      - master
-  pull_request:
-    branches:
-      - master
+on: push
```
angeloashmore, line 3: This is appreciated 🙂

.github/workflows/ci.yml
```diff
@@ -36,12 +31,34 @@ jobs:
...
 
+  suite:
+    needs: prepare
+    name: Suite (${{ matrix.os}}, Node ${{ matrix.node }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        node: [14]
```
angeloashmore, line 42: Any thoughts on including Node.js 18 in this list? Node 18 is unique since it introduces Web APIs that could affect our libraries, including `fetch`, `Response`, and `Blob`.

.github/workflows/ci.yml
```diff
@@ -52,6 +69,63 @@ jobs:
...
+        uses: actions/checkout@master
+
+      - name: Retrieve node_modules
+        uses: actions/cache@v3
+        with:
+          path: node_modules
+          key: ${{ matrix.os }}-node-v${{ matrix.node }}-deps-${{ hashFiles(format('{0}{1}', github.workspace, '/package-lock.json')) }}
+
+      - name: Overwrite TypeScript
+        run: npm install --save-dev typescript@${{ matrix.typescript }}
```
angeloashmore, line 99: Is the `--save-dev` necessary? Modifying the `package-lock.json` makes the workflow less pure since it modifies the code its testing.

We could use `--no-save` to only install TypeScript without modifying the repo.

(If you think changing `package-lock.json` doesn't make a difference, we can keep it as is.)

angeloashmore (approved):
Looks good! Just a few comments which are up to your discretion. 🙂

.github/workflows/ci.yml
lihbr: Agree! Let's do this

.github/workflows/ci.yml
lihbr: Brilliant! Yes

lihbr:
Alright
- Fixed the TypeScript version install with `--no-save`
- Tweaked the matrix to run the suite (lint, unit, build) against Node 14, 16, 18 (and I think we'll keep that flow of running against non-EOL LTS)

angeloashmore:
No comments — looks good to me!

PR #279, opened 2023-01-24
chore: deprecate `orderings` string values, resolves #269

### LINKED ISSUE #269: Deprecate `orderings` string values
### Is your feature request related to a problem? Please describe.

The `orderings` option allows for sorting query results. The client currently accepts a string in the format expected by Prismic's Rest API or an object describing a sorting.

String:
```js
{
  orderings: "[my.product.price desc]"
}
```

Object:
```js
{
  orderings: [
    { field: "my.product.price", direction: "desc" }
  ]
}
```

The second option is clearer to read, is typed via TypeScript, and shows that multiple orderings can be provided (as a result of it being an array).

The first option is easy to write incorrectly, provides no compile-time checks, and requires learning a new syntax.

### Describe the solution you'd like

Deprecate string values for the `orderings` option and only allow the object version.

Additionally, `orderings` currently accepts a single value or an array. The type should be changed to only accept an array.

### Describe alternatives you've considered

None :)

### Additional context

This request was prompted by this Prismic Community question: https://community.prismic.io/t/how-to-sort-prismic-data-by-name-or-any-other-field/11793

lihbr:
## Types of changes

## Description

Deprecate `orderings` string values and non-array values. See #269 for more details.

This is a minor breaking change for TypeScript users, while still working, type checking won't pass for deprecated values anymore, this has been reflected in the migration guide.

Resolves: #269

## Checklist:


angeloashmore:
Is it possible we could use `@deprecate` to mark `string | Ordering` as deprecated, but still accept `Ordering[]`? Perhaps using a union on the config type?

Otherwise, everything looks good!

lihbr:
> Is it possible we could use `@deprecate` to mark `string | Ordering` as deprecated, but still accept `Ordering[]`? Perhaps using a union on the config type?

I tried to do this at first yes, but no :/
Tried converting the interface to a type and extending it with a union in various ways but it either resolved to `any` or `never` :/

I fixed the conflicts, and won't merge for now in case you want to play with it yourself~ LMK

lihbr:
Again no pressure as it's a non urgent topic, but just a gentle ping ☺️ @angeloashmore

angeloashmore:
I updated the PR with the following changes:

- Restored `string | string[]` to the `orderings` parameter. Since we want to mark them as deprecated, and not removed, we need to keep the types. Removing them could break existing projects.
- Added `console.warn()` messages that print when `NODE_ENV` is `development` and a string value is detected.
- [Added a `orderings-as-array-of-objects.md` message](https://github.com/prismicio/prismic-client/blob/dee9eab5c1a1af156c087734d818c5824bdfe2f7/messages/orderings-as-array-of-objects.md) used in the `console.warn()` calls.

I think that gives us the effect we want: don't break existing projects as a result of this change, but nudge developers to update their projects.

If it looks good to you, this PR is good to merge!

angeloashmore (approved):

PR #280, opened 2023-03-27
fix: add soft support for Node.js 12

angeloashmore:
## Types of changes

## Description

This PR introduces undocumented support for Node.js 12. Although Node.js 12 is already past its end-of-life (EOL), with Node.js 14's EOL coming within the next month, some developers are still using build systems that do not recognize all Node.js 14 features.

Specifically in `@prismicio/client`, the use of the nullish coelescing operator (`??`) throws errors in older build systems (think: a Babel configuration that isn't aware of the operator).

This PR removes all use of `??` and replaces it with `||` or a nullish ternary check where appropriate.

This PR does **not** mean we officially support Node.js 12. We officially only support LTS versions of Node.js. However, because the change in the PR is simple, we can make an exception in this case.

## Checklist:

🐁

### DIFF
--- a/src/buildQueryURL.ts
+++ b/src/buildQueryURL.ts
@@ -288,7 +288,7 @@ export const buildQueryURL = (
 	// Iterate over each parameter and add it to the URL. In some cases, the
 	// parameter value needs to be transformed to fit the REST API.
 	for (const k in params) {
-		const name = (RENAMED_PARAMS[k as keyof typeof RENAMED_PARAMS] ??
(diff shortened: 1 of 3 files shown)


angeloashmore:
It's an error we've seen reported by users a few times, such as this one: https://stackoverflow.com/questions/75290373/not-sure-how-to-debug-uid-prismicdocument-uid-void-0-with-prismic-and-next

I know we can just tell users to use a supported version of Node.js or upgrade their dependencies, but this is such an easy change on our end that I feel we can make it "just work."

lihbr (approved):

lihbr:
Alright then

Issue #268, opened 2022-11-03
Provide a way to detect if a preview session is active


angeloashmore:
### Is your feature request related to a problem? Please describe.

The only way to know if a preview session is active today is to manually check for Prismic's preview cookie. Even then, only checking for presence of the cookie is not sufficient; it must be parsed to determine if a session is active.

Knowing if a preview is active is useful for displaying preview-only UI, such as a banner letting a content writer know that they are previewing draft content.

### Describe the solution you'd like

The client could provide a method that checks if a preview session is active. The client is scoped to a repository name, so the returned data could be specific to a repository. That level of specificity may be useful in sites that source content from multiple repositories.

The preview cookie is already automatically detected in `resolvePreviewURL()`, which takes into account the global scope (i.e. `document`) or a server request. The same logic could be used to support most runtimes.

Such a method could look like this:

```typescript
client.isPreviewActive() // => true or false
```

### Describe alternatives you've considered

Load the cookie, parse it, check if a specific repository name is present, return true or false. Very procedural and requires specific knowledge of Prismic's preview implementation.

### Additional context

Feature requested by @samlfair

PR #281, opened 2023-04-06
feat: add `getToolbarSrc()` helper

angeloashmore:
## Types of changes

## Description

This PR adds a `getToolbarSrc()` helper that returns a Prismic repository's Prismic Toolbar script URL.

```typescript

const toolbarSrc = prismic.getToolbarSrc("my-repo-name");
// => https://static.cdn.prismic.io/prismic.js?new=true&repo=my-repo-name
```

### Motivation

This helper will replace `@prismicio/react`'s `<PrismicToolbar>` when used in React Server Components environments; in Server Components environments, `<script>` elements should be rendered on the server, which `<PrismicToolbar>` is not able to do.

Providing `getToolbarSrc()` allows a developer to get the Prismic Toolbar's script URL and render the `<script>` using their framework's conventions. In a Next.js app, for example, a developer would use `next/script`.

```tsx
import Script from "next/script";

export default function Page() {
  const toolbarSrc = prismic.getToolbarSrc("my-repo-name");

  return (
    <main>
      <Script src={toolbarSrc} />
    </main>
  );
}
```

## Checklist:

🐋

### DIFF
--- a/src/getToolbarSrc.ts
+++ b/src/getToolbarSrc.ts
@@ -1,0 +1,50 @@
+import { PrismicError } from "./errors/PrismicError";
+
+import { isRepositoryName } from "./isRepositoryName";
+
+/**
+ * Returns the URL for a Prismic repository's Prismic Toolbar script. Use the
+ * URL to inject the script into your app.
+ *
+ * @example
+ *
+ * ```typescriptreact
+ * // In Next.js apps, use `next/script` in your `app/layout.tsx` file.
+ *
+ * import Script from "next/script";
+ * import * as prismic from '@prismicio/client'
+ *
+ * export default function RootLayout({
+ * 	children,
+ * }: {
+ * 	children: React.ReactNode,
+ * }) {
+ * 	const toolbarSrc = prismic.getToolbarSrc("my-repo");
+ *
+ * 	return (
+ * 		<html lang="en">
+ * 			<body>{children}</body>
+ * 			<Script src={toolbarSrc} />
+ * 		</html>
+ * 	);
+ * }
+ * ```
+ *
+ * @param repositoryName - The name of the Prismic repository. For example,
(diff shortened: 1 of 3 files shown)


lihbr (approved):
LGTM, don't want to add to the scope if this PR (I can do it later), but any thoughts on adding preview helpers too? + maybe we could expose `getPreviewCookie()`

Something to stop us from reimplementing this code everytime:
- https://github.com/nuxt-modules/prismic/blob/v3/src/runtime/plugin.ts#L35-L59
- https://github.com/prismicio/prismic-next/blob/cc825bd8ca36b4215880d61c7774ca9c10552f4b/src/lib/getPrismicPreviewCookie.ts
- https://github.com/prismicio/prismic-next/blob/cc825bd8ca36b4215880d61c7774ca9c10552f4b/src/PrismicPreview.tsx#L116-L149

```typescript
import { getPreviewCookie } from "./lib/getPreviewCookie";

export const isPreviewSession = (cookieJar: string): boolean => {
	const previewCookie = getPreviewCookie(cookieJar);

	if (!previewCookie) {
		return false;
	}

	const session = JSON.parse(decodeURIComponent(previewCookie));

	return Object.keys(session).some((key) => {
		return key in session &&
			typeof session[key] === 'object' &&
			session[key] !== null &&
			'preview' in session[key] &&
			session[key]
	});
};
```

angeloashmore:
> any thoughts on adding preview helpers too? + maybe we could expose `getPreviewCookie()`

Yes, let's do it! @samlfair suggested a similar thing here as well: https://github.com/prismicio/prismic-client/issues/268

Those additions are outside the scope of this PR, as you mentioned. Let's add these helpers in a separate PR:

- `isPreviewSessionActive()`
- `getPreviewCookie()`

(We may need to iterate on the names, but it's at least clear what they should do.)

PR #282, opened 2023-04-08
feat: add `asLinkAttrs()` helper

angeloashmore:
## Types of changes

## Description

This PR adds a new `asLinkAttrs()` helper. It complements the `asLink()` helper.

`asLinkAttrs()` accepts a Link field or a Prismic document and returns an object of link attributes.

```typescript
const { href, target, rel } = asLinkAttrs(myLinkField);
```

The resulting `href`, `target`, and `rel` attributes can be passed to an `<a>` element or similar link components (e.g. Next.js' `next/link` component).

```tsx
const App = () => {
  const attrs = asLinkAttrs(myLinkField);

  return <a {...attrs}>Click me!</a>;
};
```

### `rel` attribute

The `rel` value resolves to a good default: if the `href` is external (i.e. starts with a protocol like `https://`), `rel` is `"noreferrer"`.

The `rel` value can be configured using the `rel` option:

```typescript
asLinkAttrs(myLinkField, {
  rel: ({ href, isExternal, target }) => (isExternal ? "nofollow" : undefined),
});
```

The value returned by the function is exactly what is returned for the `rel` value; the automatic `noreferrer` logic described above is omitted.

Note that browsers imply the `"noopener"` value when `rel` is `"noreferrer"` _or_ `target` is `_blank`.

### Link Resolver

A Link Resolver can be provided to compute a field or document's `href`.

```typescript
asLinkAttrs(myLinkField, { linkResolver: (field) => `/${doc.uid}` });
```

Prefer using [Route Resolver](https://prismic.io/docs/route-resolver) and reserving Link Resolver only for unusual cases.

## Checklist:

🦭


test/helpers-asLink.test.ts
```diff
@@ -110,7 +110,7 @@ it("resolves a link to web field", () => {
 		url: "https://prismic.io",
 	};
 
-	expect(asLink(field, linkResolver), "https://prismic.io");
+	expect(asLink(field, linkResolver)).toBe("https://prismic.io");
```
lihbr, line 113: Weird those tests were passing 🤔

src/helpers/asLinkAttrs.ts
```diff
@@ -0,0 +1,86 @@
...
+			"target" in linkFieldOrDocument ? linkFieldOrDocument.target : undefined;
+
+		const href = asLink(linkFieldOrDocument, linkResolver);
+
+		return {
+			href: (href == null
+				? undefined
+				: href) as AsLinkAttrsReturnType<LinkResolverFunctionReturnType>["href"],
+			target,
+			rel: target === "_blank" ? "noopener noreferrer" : undefined,
```
lihbr, line 81: I'm not a fan of returning a hypothetic `rel` value, I had users ask me to have another value than `noopener noreferrer` (e.g. just `nofollow`) which ultimately lead to this value being configuration on the Vue plugin: https://prismic.io/docs/technical-reference/prismicio-vue#createprismic (see `components.linkBlankTargetRelAttribute` option

lihbr (commented):
Not sure about the `rel` value, any thought about having a similar `asImageAttrs` helper also? `asImageWidthSrcSet` and `asImagePixelDensitySrcSet` could also return `alt` value 🤔

test/helpers-asLink.test.ts
angeloashmore: Yeah, it would be nice if Vitest was smart enough to throw an error if an assertion isn't called on `expect()`. In this case, the tests passed because `expect()`'s second argument is a message, and `"https://prismic.io"` is a valid message.

src/helpers/asLinkAttrs.ts
angeloashmore: After discussing outside this thread, we decided on slightly different default `rel` behavior and the ability to override it via a function.

See the updated PR body for details.

angeloashmore:
I updated the PR with a new `rel` handling strategy. It has new default behavior (return `"noreferrer"` only when the URL is external) and allows the value to be controlled using a new `rel` option.

See the updated PR body for details.

> any thought about having a similar asImageAttrs helper also? asImageWidthSrcSet and asImagePixelDensitySrcSet could also return alt value 🤔

Makes sense to me! We will brainstorm a list of new helpers and implement them as time allows. In general, having these `asAttr()` helpers will make it easier to re-use logic across framework integrations, as well as allowing other developers to implement their own custom logic.

lihbr (approved):
LGTM!

PR #285, opened 2023-04-25
feat: add `optimize` options

angeloashmore:
## Types of changes

## Description

This PR adds a new `optimize` option to `createClient()` and all client methods that make requests to the Prismic API.

```typescript
const client = createClient({
  optimize: {
    /**
     * Determines if fetching repository metadata should be optimized. The
     * optimizations result in faster queries, less network requests, and more
     * accurate cache-busting.
     *
     * Only opt out of this optimization if you know what you are doing.
     *
     * @defaultValue `true`
     */
    repositoryRequests: true,

    /**
     * Determines if concurrent requests to the same URL should be optimized.
     * The optimizations result in less network requests and quicker responses.
     *
     * Only opt out of this optimization if you know what you are doing.
     *
     * @defaultValue `true`
     */
    concurentRequests: true,
  },
});

// `optimize` can be overridden on a per-request basis.
  optimize: {
    repositoryRequests: false,
    concurentRequests: false,
  },
});
```

All current optimizations are enabled by default.

- `repositoryRequests`: Caches successful requests to `/api/v2` internally for 5 seconds and forces external network caches to not cache `/api/v2` requests for 5 seconds.
- `concurrentRequests`: De-duplicates idential concurrent requests.

All optimizations except `repositoryRequests`'s "force external network caches to not cache" were already in place and were not configurable. This PR allows for explicitly opting out of these optimizations, which should only be done in very specific cases.

### Forcing external network caches to not cache `/api/v2`

A new optimization is introduced in this PR: requests to `/api/v2` are forcibly cache-busted by default after 5 seconds.

This is done by adding a per-request-unique URL parameter (`x-valid-until`) containing an epoch timestamp produced at the time of the request. The timestamp represents the time at which the request is no longer valid.

```
https://example-prismic-repo.cdn.prismic.io/api/v2?x-valid-until=1682556195000
```

The 5 second lifespan is not guaranteed. It is calculated by taking the timestamp at the time of the request, rounding down to the nearest 5 seconds, then adding 5 seconds. Treat it as an unguaranteed optimization that may not always be respected.

Only `/api/v2` requests are optimized. All `/api/v2/documents/search` requests are inherently cache-busted by the `ref` parameter.

The parameter is ignored within the concurrent requests job queue when `optimize.concurrentRequests` is enabled.

This new optimization was prompted by the need to forcibly opt out of [Next.js's new aggressive App Router `fetch()` caching](https://beta.nextjs.org/docs/data-fetching/caching).

### To discuss

1. **Is this PR over-engineered?**
   We technically only need to add the `x-valid-until` parameter to every network request with the option to opt out. The other `optimize` options are nice-to-haves to cover edge cases.

2. **Is `x-valid-until` the correct parameter name?**

3. **Is `optimize` the correct option name?**
   It was inspired by Vite's `optimizeDeps` option.

## Checklist:

🦥

### DIFF
--- a/src/createClient.ts
+++ b/src/createClient.ts
@@ -50,6 +50,13 @@ export const REPOSITORY_CACHE_TTL = 5000;
 export const GET_ALL_QUERY_DELAY = 500;
 
 /**
+ * The URL search parameter name used for optimizing `/api/v2` requests. The
+ * value of the parameter effectively cache-busts the request by providing a
+ * unique value on every request.
+ */
+const OPTIMIZE_REPOSITORY_REQUESTS_TIMESTAMP_PARAMETER_NAME = "x-optimize-ts";
+
+/**
  * Extracts one or more Prismic document types that match a given Prismic
  * document type. If no matches are found, no extraction is performed and the
  * union of all provided Prismic document types are returned.
@@ -257,7 +264,7 @@ export type ClientConfig = {
 	 * Node.js, this function must be provided.
 	 */
 	fetch?: FetchLike;
-};
+} & OptimizeParams;
 
 /**
  * Parameters for any client method that use `fetch()`. Only a subset of
@@ -271,6 +278,36 @@ type FetchParams = {
 	 * {@link https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal}
 	 */
 	signal?: AbortSignalLike;
+};
+
+/**
+ * Parameters that determine if or how the client is optimized.
+ */
+type OptimizeParams = {
+	/**
+	 * Options that determine if or how the client is optimized.
+	 */
+	optimize?: {
+		/**
+		 * Determines if fetching repository metadata should be optimized. The
+		 * optimizations result in faster queries, less network requests, and more
+		 * accurate cache-busting.
+		 *
+		 * Only opt out of this optimization if you know what you are doing.
+		 *
+		 * @defaultValue `true`
+		 */
+		repositoryRequests?: boolean;
+
+		/**
+		 * Determines if concurrent requests to the same URL should be optimized. The optimizations result in less network requests and quicker responses.
+		 *
+		 * Only opt out of this optimization if you know what you are doing.
+		 *
+		 * @defaultValue `true`
+		 */
+		concurentRequests?: boolean;
+	};
 };
 
 /**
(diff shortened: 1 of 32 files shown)


angeloashmore:
I think there's a bug somewhere in the query-level `optimize` overrides. The latest commit's tests show the inconsistencies.

I'll do some digging to figure out what's going on.

src/createClient.ts
```diff
@@ -1183,6 +1241,28 @@ export class Client<TDocuments extends PrismicDocument = PrismicDocument> {
...
+			//
+			// This fingerprint works as a second "cache" that
+			// works outside `getRepositoryCache()`. This is useful
+			// when multiple client instances exist with an
+			// external cache; all clients can share the same URL
+			// cache key for the duration of REPOSITORY_CACHE_TTL.
+			url.searchParams.set(
+				OPTIMIZE_REPOSITORY_REQUESTS_VALID_UNTIL_PARAMETER_NAME,
+				(
+					Math.ceil(Math.floor(Date.now() / REPOSITORY_CACHE_TTL)) *
```
lihbr, line 1260: I think I'm missing something, but why do we need to ceil something that was floored just before?

lihbr (approved):
LGTM, maybe one small typo? See comment

src/createClient.ts
lihbr: I don't see where we add the `RESPOSITORY_CACHE_TTL` to the 5s-rounded TS here 🤔

This, `Date.now() >= Math.ceil(Math.floor(Date.now() / REPOSITORY_CACHE_TTL)) * REPOSITORY_CACHE_TTL`, seems to be always true

src/createClient.ts
```diff
@@ -1183,6 +1241,28 @@ export class Client<TDocuments extends PrismicDocument = PrismicDocument> {
...
+					REPOSITORY_CACHE_TTL
```
lihbr, line 1261:
```suggestion
					Math.floor(Date.now() / REPOSITORY_CACHE_TTL) *
					REPOSITORY_CACHE_TTL + REPOSITORY_CACHE_TTL
```

src/createClient.ts
angeloashmore: Haha that's for catching that. The formula was different before and required `ceil` + `floor`, but it's no longer required after some simplification.

Your `+ REPOSITORY_CACHE_TTL` is correct; I must have copied it wrong from my sandbox.

Issue #283, opened 2023-04-11
Align helpers to `fn(field, options)` interface


lihbr:
### Is your feature request related to a problem? Please describe.

Right now helpers interfaces aren't consistent:
- Some are `fn(field, option1, option2, ...)`
- Other are `fn(field, options)`

### Describe the solution you'd like

Aligning all helpers to `fn(field, options)` interface, deprecated but still supporting `fn(field, option1, option2, ...)` interface.

This mainly impact `asLink` and `asHTML` helpers.

### Describe alternatives you've considered

Doing nothing -> inconsistent interfaces

### Additional context

(internal): https://prismic-team.slack.com/archives/D01UP4FN3PX/p1681156796335669

PR #288, opened 2023-05-02
refactor(helpers): standardize helpers interface

lihbr:
## Types of changes

## Description

Resolves: #283

## Checklist:

### DIFF
--- a/src/helpers/asHTML.ts
+++ b/src/helpers/asHTML.ts
@@ -204,40 +204,85 @@ const wrapMapSerializerWithStringChildren = (
 };
 
 /**
+ * Configuration that determines the output of `asHTML()`.
+ */
+type AsHTMLConfig = {
+	/**
+	 * An optional link resolver function to resolve links.
+	 * Without it you're expected to use the `routes` options from the API.
+	 */
+	linkResolver?: LinkResolverFunction<string | null | undefined> | null;
+
+	/**
+	 * An optional Rich Text Serializer, unhandled cases will fallback to the default serializer
+	 */
+	htmlRichTextSerializer?: HTMLRichTextSerializer | null;
+};
+
+// TODO: Remove when we remove support for deprecated tuple-style configuration.
+/**
+ * @deprecated Use object-style configuration instead.
+ */
+type AsHTMLDeprecatedTupleConfig = [
+	linkResolver?: LinkResolverFunction<string | null | undefined> | null,
+	htmlRichTextSerializer?: HTMLRichTextSerializer | null,
+];
+
+/**
  * The return type of `asHTML()`.
  */
 type AsHTMLReturnType<Field extends RichTextField | null | undefined> =
 	Field extends RichTextField ? string : null;
 
 /**
- * Serializes a Rich Text or Title field to an HTML string
+ * Serializes a Rich Text or Title field to an HTML string.
  *
  * @param richTextField - A Rich Text or Title field from Prismic
- * @param linkResolver - An optional link resolver function to resolve links,
- *   without it you're expected to use the `routes` options from the API
- * @param htmlRichTextSerializer - An optional Rich Text Serializer, unhandled cases will fallback
- *   to the default serializer
+ * @param configObjectOrTuple - Configuration that determines the output of `asHTML()`
  *
  * @returns HTML equivalent of the provided Rich Text or Title field
  * @see Templating Rich Text and title fields from Prismic {@link https://prismic.io/docs/template-content-vanilla-javascript#rich-text-and-title}
  */
 export const asHTML = <Field extends RichTextField | null | undefined>(
 	richTextField: Field,
-	linkResolver?: LinkResolverFunction<string | null | undefined> | null,
-	htmlRichTextSerializer?: HTMLRichTextSerializer | null,
+	// TODO: Rename to `config` when we remove support for deprecated tuple-style configuration.
+	...configObjectOrTuple: [config?: AsHTMLConfig] | AsHTMLDeprecatedTupleConfig
 ): AsHTMLReturnType<Field> => {
 	if (richTextField) {
+		// TODO: Remove when we remove support for deprecated tuple-style configuration.
+		const [configObjectOrLinkResolver, maybeHTMLRichTextSerializer] =
+			configObjectOrTuple;
+		let config: AsHTMLConfig;
+		if (
+			typeof configObjectOrLinkResolver === "function" ||
+			configObjectOrLinkResolver == null
+		) {
+			config = {
+				linkResolver: configObjectOrLinkResolver,
+				htmlRichTextSerializer: maybeHTMLRichTextSerializer,
+			};
+		} else {
+			config = { ...configObjectOrLinkResolver };
+		}
+
 		let serializer: RichTextFunctionSerializer<string>;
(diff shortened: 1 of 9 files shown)


src/helpers/asHTML.ts
```diff
@@ -203,41 +203,86 @@ const wrapMapSerializerWithStringChildren = (
...
 
+/**
+ * Configuration that determines the output of `asHTML()`.
+ */
+type AsHTMLConfig = {
+	/**
+	 * An optional link resolver function to resolve links.
+	 * Without it you're expected to use the `routes` options from the API.
+	 */
+	linkResolver?: LinkResolverFunction<string | null | undefined> | null;
```
angeloashmore, line 214: I believe we can remove the `<string | null | undefined>` as that is now the default. Could you confirm?

src/helpers/asHTML.ts
```diff
@@ -203,41 +203,86 @@ const wrapMapSerializerWithStringChildren = (
...
+	 */
+	htmlRichTextSerializer?: HTMLRichTextSerializer | null;
+};
+
+// TODO: Remove when we remove support for deprecated tuple-style configuration.
+/**
+ * @deprecated Use object-style configuration instead.
+ */
+type AsHTMLDeprecatedTupleConfig = [
+	linkResolver?: LinkResolverFunction<string | null | undefined> | null,
```
angeloashmore, line 227: Same as above re: `LinkResolverFunction`

src/helpers/asHTML.ts
```diff
@@ -203,41 +203,86 @@ const wrapMapSerializerWithStringChildren = (
...
  *
  * @returns HTML equivalent of the provided Rich Text or Title field
  * @see Templating Rich Text and title fields from Prismic {@link https://prismic.io/docs/template-content-vanilla-javascript#rich-text-and-title}
  */
 export const asHTML = <Field extends RichTextField | null | undefined>(
 	richTextField: Field,
-	linkResolver?: LinkResolverFunction<string | null | undefined> | null,
-	htmlRichTextSerializer?: HTMLRichTextSerializer | null,
+	// TODO: Rename to `config` when we remove support for deprecated tuple-style configuration.
+	...configObjectOrTuple: [config?: AsHTMLConfig] | AsHTMLDeprecatedTupleConfig
```
angeloashmore, line 249: Since the argument name is public in `tsserver`, could we rename `configObjectOrTuple` to `config`? Accepting a tuple is an implementation detail that users don't need to know.

To make sure the public signature looks correct, we might benefit from using a function overload in this case. That way users will see the two versions of the function: the object config and the multiple parameter version.

src/helpers/asHTML.ts
```diff
@@ -203,41 +203,86 @@ const wrapMapSerializerWithStringChildren = (
...
+			typeof configObjectOrLinkResolver === "function" ||
+			configObjectOrLinkResolver == null
+		) {
+			config = {
+				linkResolver: configObjectOrLinkResolver,
+				htmlRichTextSerializer: maybeHTMLRichTextSerializer,
+			};
+		} else {
+			config = { ...configObjectOrLinkResolver };
+		}
```
angeloashmore, line 266: This seems a bit complex. Is there any way to simplify this into a single `config` reassignment?

```ts
config = {
  linkResolver: ...,
  htmlRichTextSerializer: ...,
}
```

src/helpers/asLink.ts
```diff
@@ -60,7 +81,10 @@ export const asLink = <
 		| undefined,
 >(
 	linkFieldOrDocument: Field,
-	linkResolver?: LinkResolverFunction<LinkResolverFunctionReturnType> | null,
+	// TODO: Rename to `config` when we remove support for deprecated tuple-style configuration.
+	...configObjectOrTuple:
+		| [config?: AsLinkConfig<LinkResolverFunctionReturnType>]
+		| AsLinkDeprecatedTupleConfig<LinkResolverFunctionReturnType>
```
angeloashmore, line 87: Same `config` comments as left on `asHTML()`.

src/helpers/asText.ts
```diff
@@ -2,27 +2,60 @@ import { asText as baseAsText } from "@prismicio/richtext";
...
+ * @param configObjectOrTuple - Configuration that determines the output of `asText()`
  *
  * @returns Plain text equivalent of the provided Rich Text or Title field
  * @see Templating Rich Text and title fields from Prismic {@link https://prismic.io/docs/template-content-vanilla-javascript#rich-text-and-title}
  */
 export const asText = <Field extends RichTextField | null | undefined>(
 	richTextField: Field,
-	separator?: string,
+	// TODO: Rename to `config` when we remove support for deprecated tuple-style configuration.
+	...configObjectOrTuple: [config?: AsTextConfig] | AsTextDeprecatedTupleConfig
```
angeloashmore, line 41: Same comments as left on `asHTML()` (I'm just repeating the comment so it's not forgotten 🙂)

test/helpers-asHTML.test.ts
```diff
@@ -8,23 +8,54 @@ import { richTextFixture } from "./__fixtures__/richText";
 import { RichTextField, asHTML } from "../src";
 
 it("serializes with default serializer", () => {
-	expect(asHTML(richTextFixture.en, linkResolver)).toMatchSnapshot();
+	expect(asHTML(richTextFixture.en, { linkResolver })).toMatchSnapshot();
+
+	// TODO: Remove when we remove support for deprecated tuple-style configuration.
+	expect(asHTML(richTextFixture.en, linkResolver)).toBe(
+		asHTML(richTextFixture.en, { linkResolver }),
+	);
```
angeloashmore, line 16: This is clever 🙂

angeloashmore (changes requested):
Looks good! I think there are a few places we could tidy up, but I'll leave it up to you to decide if it's possible.

src/helpers/asHTML.ts
lihbr: Indeed!

src/helpers/asHTML.ts
lihbr: TIL you can now do fancier things with overload, I'll let you checks that it works as expected on your end but it should also show users they are using the deprecated overload explicitely.

src/helpers/asHTML.ts
lihbr: I could refactor to a ternary, but I don't see the benefit in readability tbh 🤔

lihbr:
Alright, got the overload to work, it seems to work better than in my memory, we should now get clean overloads, thanks for suggesting to dig into that.

<details>
<summary>Overloads on VSCode</summary>
<br />

image
image
image
image

</details>

May I just ask you to double-check that everything looks alright on Vim too? If yes, then feel free to hit merge ☺️

src/helpers/asHTML.ts
angeloashmore: No worries! If you don't see a benefit or better way of doing it, let's keep it as is.

angeloashmore:
Looks perfect in Vim! Nice work on the overloads. Being able to use `@deprecated` is a nice touch. I just wish `tsserver` would display the deprecation message, but this is still good.

angeloashmore (approved):

PR #289, opened 2023-05-05
feat: add `fetchOptions` parameter to control `fetch()`

angeloashmore:
## Types of changes

## Description

This PR adds a new `fetchOptions` parameter to `createClient()` and all client methods that make network requests.

The parameter allows for customizing how `fetch()` is called.

```typescript
// Set `fetch()` options for every request.
const client = createClient("example-prismic-repo", {
  fetchOptions: {
    cache: "force-cache",
  },
});

// Set or override `fetch()` options for individual queries.
  fetchOptions: {
    cache: "no-store",
  },
});
```

### Supported options

The TypeScript types support a subset of the global `RequestInit` type:

- [`headers`](https://developer.mozilla.org/en-US/docs/Web/API/Request/headers) (only plain objects are supported)
- [`cache`](https://developer.mozilla.org/en-US/docs/Web/API/Request/cache)
- [`signal`](https://developer.mozilla.org/en-US/docs/Web/API/Request/signal)

These three parameters have known real-world use cases, while others, like [`credentials`](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials), currently do not.

If we learn that other parameters are needed, we can add support for them in the future.

### Adding support for project-specific options

`@prismicio/client`'s type for `fetchOptions` can be augmented on a per-project basis using [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation).

This is useful if a project needs to add support for a `RequestInit` property that is not supported by default. It is also useful if projects have project-specific `RequestInit` features that need to be added, like [Next.js' `next` parameter](https://nextjs.org/docs/app/api-reference/functions/fetch#optionsnextrevalidate).

```typescript
// Adds `credentials` support.
declare module "@prismicio/client" {
  interface RequestInitLike {
    credentials?: RequestInit["credentials"];
  }
}

const client = createClient("example-prismic-repo", {
  fetchOptions: {
    credentials: "same-origin",
  },
});
```

### Deprecating `signal` parameter in favor of `fetchOptions.signal`

The `signal` parameter has been replaced by the more apt `fetchOptions.signal`.

```typescript
const controller = new AbortSignal();
const client = createClient("example-prismic-repo");

// New
await client.getByUID("page", "home", {
  fetchOptions: {
    signal: controller.signal,
  },
});

// Old
await client.getByUID("page", "home", {
  signal: controller.signal,
});
```

`signal` will continue to be supported, but has been marked as deprecated. Please migrate to `fetchOptions.signal`.

## Checklist:

🐕

### DIFF
--- a/src/createClient.ts
+++ b/src/createClient.ts
@@ -97,16 +97,23 @@ export type FetchLike = (
 export type AbortSignalLike = any;
 
 /**
- * The minimum required properties from RequestInit.
+ * A subset of RequestInit properties to configure a `fetch()` request.
  */
-export interface RequestInitLike {
+// Only options relevant to the client are included. Extending from the full
+// RequestInit would cause issues, such as accepting Header objects.
+//
+// An interface is used to allow other libraries to augment the type with
+// environment-specific types.
+export interface RequestInitLike extends Pick<RequestInit, "cache"> {
+	/**
+	 * An object literal to set the `fetch()` request's headers.
+	 */
 	headers?: Record<string, string>;
 
 	/**
-	 * An object that allows you to abort a `fetch()` request if needed via an
-	 * `AbortController` object
+	 * An AbortSignal to set the `fetch()` request's signal.
 	 *
-	 * {@link https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal}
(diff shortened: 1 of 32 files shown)


lihbr (approved):
Looking sharp 👌

PR #299, opened 2023-05-20
fix: support thunk values for `fetchOptions` parameter

angeloashmore:
**DO NOT MERGE: This PR may not be necessary. Further exploration is needed.**

## Types of changes

## Description

> **Note**: This PR is technically a new feature. Considering v7.0.0 was just released and `fetchOptions` is not yet widely used, I think we can considering this addition a patch in the spirit of: "this should have been included in the original version."

This PR adds the ability to provide thunks to the `fetchOptions` parameter. Before this PR, only plain objects could be provided.

```typescript
import { createClient } from "@prismicio/client";

const client = createClient("example-prismic-repo", {
  // `url` is the network request's URL, allowing for customization per network request.
  // `init` is the calling method's resolved `fetchOptions`, if given.
  fetchOptions: (url, init) => {
    return {
      cache: "no-store",
      ...init,
    };
  },
});

// The client's default `fetchOptions` take priority.
  // `url` is the network request's URL, allowing for customization per network request.
  // `init` is undefined.
  fetchOptions: (url, init) => {
    return {
      cache: "no-store",
    };
  },
});
```

Note that the client's `fetchOptions` take priority. This means that in order to override `fetchOptions` from a method, the client method **must** use a thunk and merge the `init` value.

### Use case

Using a thunk is needed to selectively apply a Next.js `fetch()` tag to a request. We need to tag `/api/v2` requests differently from `/api/v2/documents/search` requests.

With the appropriate tags, we can selectively revalidate (i.e. clear) the cache programmatically via `next/cache`'s `revalidateTag()` function.

#### Example

A client:

```typescript
const client = prismic.createClient(repositoryName, {
  routes,
  fetchOptions: (url, init) => {
    return {
      cache: "force-cache",
      ...init,
      next: {
        ...init?.next,
        tags: [
          "prismic",
          /\/documents\/search\/?/.test(url)
            ? "prismic-query"
            : "prismic-repository",
          ...(init?.next?.tags || []),
        ],
      },
    };
  },
  ...config,
});
```

In a Route Handler to revalidate `/api/v2`:

```typescript
// src/app/revalidate/route.ts

import { NextResponse } from "next/server";
import { revalidateTag } from "next/cache";

export async function GET() {
  revalidateTag("prismic-repository");

  return NextResponse.json({ revalidated: true, now: Date.now() });
}
```

## Checklist:

🐻‍❄️


lihbr (approved):
LGTM, all good for the patch release~

angeloashmore:
Thanks for the review! 🙂

I'm going to close this PR since it is no longer needed for Next.js apps. We discovered a simpler way to manage the cache that does not require a thunk.

```typescript
const client = createClient("example-prismic-repo", {
  // `url` is the network request's URL, allowing for customization per network request.
  // `init` is the calling method's resolved `fetchOptions`, if given.
  fetchOptions: {
    cache: process.env.NODE_ENV === "production" ? "force-cache" : "no-store",
    next: { tags: ["prismic"] },
  },
});
```

PR #302, opened 2023-05-31
feat: add `unstable_mapSliceZone()`

angeloashmore:
## Types of changes

## Description

This PR adds an experimental `unstable_mapSliceZone()` helper. The helper maps a Slice Zone's Slices using a set of mapping functions, one for each type of a Slice Zone's Slice.

Developers often need to transform or augment Slices when building relatively complex websites. We currently do not offer an official solution for this need, thus developers come up with their own non-standard solutions.

This function is primarily recommended for use on the server where Slices can be augmented, reshaped, or trimmed-down.

The solution provided in this PR is general, flexible, and tested.

### Example

The following example reshapes Code Block Slices (id: `code_block`) by replacing it with an object containing a single `codeHTML` property. The property contains HTML produced from a `highlight()` function, which could be a function that calls a syntax highlighting library, like Shiki.

```typescript
async function highlight(code: string, language: string): Promise<string> {
  // Process the code and return HTML.
}

const mappedSliceZone = await unstable_mapSliceZone(page.data.slices, {
  code_block: async ({ slice }) => ({
    codeHTML: await highlight(slice.primary.code, slice.primary.language),
  }),
});
```

Any occurance of a Code Block Slice will be replaced with the mapped version. Note how the function provided to the `code_block` property receives the `slice` object. Other data is included, such as `slices`, and `index`.

A third parameter (`context`) can be provided to `unstable_mapSliceZone()`, whose contents will be passed to the function in the `context` property. The context object is most helpful when mapping functions are defined outside the `unstable_mapSliceZone()` calling scope, such as in a different module.

```typescript
const mappedSliceZone = await unstable_mapSliceZone(
  page.data.slices,
  {
    code_block: async ({ slice, context }) => ({
      codeHTML: await highlight(slice.primary.code, slice.primary.language),
    }),
  },
  context
);
```

### Unstable

This helper is marked as `unstable_` in its function name and `@experimental` in its TSDocs. It is being added as a way for developers to test it in production and provide feedback before being marked as stable.

## Checklist:

🦦


src/helpers/unstable_mapSliceZone.ts
```diff
@@ -0,0 +1,266 @@
...
+	: Slice extends SliceLikeGraphQL
+	? Slice["type"]
+	: never;
+
+/**
+ * The minimum required properties to represent a Prismic Slice from the Prismic
+ * Rest API V2 for the `<SliceZone>` component.
+ *
+ * If using Prismic's Rest API V2, use the `Slice` export from
+ * `@prismicio/types` for a full interface.
```
lihbr, line 25:
```suggestion
 * The minimum required properties to represent a Prismic Slice from the Prismic
 * Rest API V2 for the `mapSliceZone` helper.
 *
 * If using Prismic's Rest API V2, use the `Slice` export from
 * `@prismicio/client` for a full interface.
```

src/helpers/unstable_mapSliceZone.ts
```diff
@@ -0,0 +1,266 @@
...
+ * @typeParam SliceType - Type name of the Slice.
+ */
+export type SliceLikeRestV2<SliceType extends string = string> = {
+	slice_type: Slice<SliceType>["slice_type"];
+	id?: string;
+};
+
+/**
+ * The minimum required properties to represent a Prismic Slice from the Prismic
+ * GraphQL API for the `<SliceZone>` component.
```
lihbr, line 36:
```suggestion
 * The minimum required properties to represent a Prismic Slice from the Prismic
 * GraphQL API for the `mapSliceZone` helper.
```

src/helpers/unstable_mapSliceZone.ts
```diff
@@ -0,0 +1,266 @@
...
+export type SliceLikeGraphQL<SliceType extends string = string> = {
+	type: Slice<SliceType>["slice_type"];
+};
+
+/**
+ * The minimum required properties to represent a Prismic Slice for the
+ * `<SliceZone>` component.
+ *
+ * If using Prismic's Rest API V2, use the `Slice` export from
+ * `@prismicio/types` for a full interface.
```
lihbr, line 49:
```suggestion
 * The minimum required properties to represent a Prismic Slice for the
 * `mapSliceZone` helper.
 *
 * If using Prismic's Rest API V2, use the `Slice` export from
 * `@prismicio/client` for a full interface.
```

src/helpers/unstable_mapSliceZone.ts
```diff
@@ -0,0 +1,266 @@
...
+export type SliceLike<SliceType extends string = string> =
+	| SliceLikeRestV2<SliceType>
+	| SliceLikeGraphQL<SliceType>;
+
+/**
+ * A looser version of the `SliceZone` type from `@prismicio/types` using
+ * `SliceLike`.
+ *
+ * If using Prismic's Rest API V2, use the `SliceZone` export from
+ * `@prismicio/types` for the full type.
```
lihbr, line 62:
```suggestion
 * A looser version of the `SliceZone` type from `@prismicio/client` using
 * `SliceLike`.
 *
 * If using Prismic's Rest API V2, use the `SliceZone` export from
 * `@prismicio/client` for the full type.
```

src/helpers/unstable_mapSliceZone.ts
```diff
@@ -0,0 +1,266 @@
...
+	__mapped: true;
+};
+
+/**
+ * React props for a component rendering content from a Prismic Slice using the
+ * `<SliceZone>` component.
+ *
+ * @typeParam TSlice - The Slice passed as a prop.
+ * @typeParam TContext - Arbitrary data passed to `<SliceZone>` and made
+ *   available to all Slice components.
```
lihbr, line 88:
```suggestion
 * Config for a function mapping content from a Prismic Slice using the
 * `mapSliceZone` helper.
 *
 * @typeParam TSlice - The Slice passed as a prop.
 * @typeParam TContext - Arbitrary data passed to `mapSliceZone` and made
 *   available to all Slice mappers.
```

src/helpers/unstable_mapSliceZone.ts
```diff
@@ -0,0 +1,266 @@
...
+ */
+export type SliceComponentProps<
```
lihbr, line 90:
```suggestion
export type SliceMapperConfig<
```

src/helpers/unstable_mapSliceZone.ts
```diff
@@ -0,0 +1,266 @@
...
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	TSlice extends SliceLike = any,
+	TContext = unknown,
+> = {
+	/**
+	 * Slice data for this component.
```
lihbr, line 96:
```suggestion
	 * Slice data for this mapper.
```

src/helpers/unstable_mapSliceZone.ts
```diff
@@ -0,0 +1,266 @@
...
+	 */
+	// TODO: We have to keep this list of Slices general due to circular
+	// reference limtiations. If we had another generic to determine the full
+	// union of Slice types, it would include TSlice. This causes TypeScript to
+	// throw a compilation error.
+	slices: SliceZoneLike<SliceLike>;
+
+	/**
+	 * Arbitrary data passed to `<SliceZone>` and made available to all Slice
+	 * components.
```
lihbr, line 116:
```suggestion
	 * Arbitrary data passed to `mapSliceZone` and made available to all Slice
	 * mappers.
```

src/helpers/unstable_mapSliceZone.ts
```diff
@@ -0,0 +1,266 @@
...
+
+			let result = await mapper(mapperArgs);
+
+			if (
+				mapper.length < 1 &&
+				(typeof result === "function" ||
+					(typeof result === "object" && "default" in result))
+			) {
+				result = "default" in result ? result.default : result;
+				result = await result(mapperArgs);
```
lihbr, line 247: I'd maybe rename the first `result` to something like `resultOrMapperModule` so that the following line can read better.

src/helpers/unstable_mapSliceZone.ts
```diff
@@ -0,0 +1,266 @@
...
+ * A set of properties that identify a Slice as having been mapped.
+ * `<SliceZone>` uses these properties internally to determine how props are
+ * provided to a Slice's component.
+ */
+type MappedSliceLike = {
+	/**
+	 * If `true`, this Slice has been modified from its original value using a
+	 * mapper.
+	 */
+	__mapped: true;
```
lihbr, line 79: Can we `@internal` it so it's more likely to be hidden from intellisense?

test/helpers-unstable_mapSliceZone.test.ts
```diff
@@ -0,0 +1,188 @@
...
+		{ __mapped: true, id: undefined, slice_type: model1.id, foo: "bar" },
+		{ __mapped: true, id: undefined, slice_type: model2.id, baz: "qux" },
+	]);
+});
+
+it("supports mapping functions that return undefined", async (ctx) => {
+	const { sliceZone, model1, model2 } = generateTestData(ctx);
+
+	const actual = await unstable_mapSliceZone(sliceZone, {
+		[model1.id]: () => void 0,
```
lihbr, line 53: You're that kind of guy now, I see 🕶️

lihbr (changes requested):
Made some suggestions, but actually, are you also using this PR to export and reuse looser types for `<SliceZone />` component implementations? 🤔 

I'm not sure about the `_unstable` prefix, is it that unstable, as in, is there still chances for us to get rid of it? If yes, then sure, let's keep it, otherwise I'd consider dropping it.

src/helpers/unstable_mapSliceZone.ts
angeloashmore: I prefer `result` here since we ultimately will be using `result` as the final mapper function return value. `result` is used later in the function to spread into an object, which should never be a module.

However, I added comments around `const result` to explain exactly what is happening at that stage. I think that should accomplish what your rename suggestion is doing.

What do you think?

src/helpers/unstable_mapSliceZone.ts
angeloashmore: I'm not sure if we should mark it as `@internal`. We need to expose it so consumers of the Slice Zone can use it. It is a public way to identify if a Slice has been mapped.

We initially only plan to use it within `<SliceZone>` so we can determine how to pass a component its props, but there may be other use cases for mapping a Slice Zone.

test/helpers-unstable_mapSliceZone.test.ts
angeloashmore: `void 0` is just so much cooler than `undefined`

src/helpers/unstable_mapSliceZone.ts
lihbr: Works for me!

test/helpers-unstable_mapSliceZone.test.ts
lihbr: What about `[][0]` then? 🙃

angeloashmore:
> Made some suggestions, but actually, are you also using this PR to export and reuse looser types for `<SliceZone />` component implementations? 🤔

Thanks for catching the clearly incorrect TSDocs! Can you tell I copied it over from the initial `@prismicio/react` implementation? 😅

The types should not have been exported. I also changed the loose `SliceLike` type to `Slice`, but kept `SliceGraphQLLike` since this function should support the GraphQL API where we don't always know the shape of the object.

> I'm not sure about the `_unstable` prefix, is it that unstable, as in, is there still chances for us to get rid of it? If yes, then sure, let's keep it, otherwise I'd consider dropping it.

`unstable_` is a React convention to mark a function as volatile. Its implementation or APIs may change and should be considered unfinished. Once we determine that this function works well and is useful, we can remove the `unstable_` prefix and deem it stable.

This function designed for a common, but advanced, use case, so I'm not 100% sure it should be treated the same as our other helpers, like `asDate()` and `asLink()`. For now, we can safely treat it as an experiment with no downsides.

What do you think though? Should we just commit to adding it to `@prismicio/client` as a stable helper?

angeloashmore:
The types are failing due to the changes in #304. Once we publish #304, which I'll include as part of publishing this PR, the type errors should go away.

PR #304, opened 2023-06-01
fix: type `PrismicDocument.*_publication_date` as `TimestampField<"filled">`

angeloashmore:
## Types of changes

## Description

This PR changes the following fields from `string` to `TimestampField<"filled">`:

- `PrismicDocument.first_publication_date`
- `PrismicDocument.last_publication_date`

Retyping the properties makes them compatible with `asDate()` when type-checked.

Requested by @chamois-d-or

#### Will this have negative downstream effects?

Changing the types makes the properties more specific, which shouldn't affect user code that accepted one of the properties. It could, however, cause issues in edge cases where user code was overriding the properties.

#### Is there a better way to install `@prismicio/client` (i.e. itself) as `@prismicio/mock`'s peer dependency?

`@prismicio/mock` was updated with a peer dependency on `@prismicio/client`. npm was installing the published version alongside `mock`, which did not include the updated types.

Because this PR changes `PrismicDocument` types, it caused type incompatibilities between `@prismicio/mock` and `src/` version of `@prismicio/client`.

I was able to solve the issue by pointing the `@prismicio/client` dependency on itself via `"@prismicio/client": "."`, which is a very odd setup.

## Checklist:

🦒


lihbr (approved):
LGTM!

angeloashmore:
@lihbr Could you take one more look? The most recent commit ([see here](https://github.com/prismicio/prismic-client/pull/304/commits/4ce1dea51e5bed4ce2eda802a6f941fcf2e34b1b)) shows how I removed `"@prismicio/client": "."` from `package.json`.

Doing so required using `@ts-expect-error`, which should be removable once we merge and publish this version.

lihbr (approved):
All good dropping Node 14!

For the `@ts-expect-error`, maybe let's just not add them, ignore the CI & publish manually (bypassing part of the release script)?

PR #319, opened 2023-08-23
feat: automatically retry rate-limited requests

angeloashmore:
## Types of changes

## Description

This PR adds automatic retry support when Prismic Rest API V2 returns a "429: Too many requests" error response.

The 429 error response can be returned if the API rate limit is exceeded. As of this PR, **the API has a limit of 200 requests per second**.

`@prismicio/client` will automatically retry those requests after enough time has ellapsed to avoid the rate limit. The retry will be transparent from client code, except for the extra wait time.

Code using `@prismicio/client` does not need to change to take advantage of the automatic retry feature.

## Checklist:

🦔

### DIFF
--- a/src/createClient.ts
+++ b/src/createClient.ts
@@ -52,6 +52,15 @@ export const REPOSITORY_CACHE_TTL = 5000;
 export const GET_ALL_QUERY_DELAY = 500;
 
 /**
+ * The number of milliseconds to wait before retrying a rate-limited `fetch()`
+ * request (429 response code). The default value is only used if the response
+ * does not include a `retry-after` header.
+ *
+ * The API allows up to 200 requests per second.
+ */
+const DEFUALT_RETRY_AFTER_MS = 1000;
+
+/**
  * Extracts one or more Prismic document types that match a given Prismic
  * document type. If no matches are found, no extraction is performed and the
  * union of all provided Prismic document types are returned.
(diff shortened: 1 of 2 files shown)


src/createClient.ts
```diff
@@ -51,6 +51,15 @@ export const REPOSITORY_CACHE_TTL = 5000;
...
 export const GET_ALL_QUERY_DELAY = 500;
 
+/**
+ * The default number of milliseconds to wait before retrying a rate-limited
+ * `fetch()` request (429 response code). The default value is only used if the
+ * response does not include a `retry-after` header.
+ *
+ * The API allows up to 200 requests per second.
+ */
+const DEFUALT_RETRY_AFTER_MS = 1000;
```
lihbr, line 61: Couldn't/doesn't the API return proper rate limit headers consistently so we can base the timeout on those?

In all fairness though there's no real standard apart from this draft: https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-07.html, been using GitHub's on my services~

lihbr (approved):
LGTM!

src/createClient.ts
angeloashmore: In the planned update, the API will only return an additional `Retry-After` header. It will always return `0`, but I asked the platform team if it could return the number of milliseconds until the request can be retried successfully (as the header's name implies).

See the internal Slack discussion here: https://prismic-team.slack.com/archives/C030DCL4JNR/p1692824110830889

In the current code, that means the request will be retried immediately since it uses `0` as `setTimeout()`'s delay. `DEFAULT_RETRY_AFTER_MS` is only used if the header is missing for some reason.

If we cannot get `Retry-After` to have millisecond precision, we should at least update the value to `1` (seconds) or `1000` (milliseconds) so that the client is properly retry.

angeloashmore: FYI, the API now returns the number of milliseconds until the rate limit is reset via the `Retry-After` header. 🙂

PR #318, opened 2023-08-21
feat: add `@prismicio/client/richtext` entry

angeloashmore:
## Types of changes

- [ ] Chore (a non-breaking change which is related to package maintenance)
- [ ] Bug fix (a non-breaking change which fixes an issue)
- [x] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

This PR adds a new `@prismicio/client/richtext` entry.

It ports all of `@prismicio/richtext` to `@prismicio/client` as an isolated entry. An isolated entry was selected over exporting it adjacent to all other helpers to minimize accidental usage; rich text helpers are low-level functions to be used primarily in high-level framework SDKs, like `@prismicio/react` and `@prismicio/vue`.

`@prismicio/richtext` will be deprecated once this PR is published. All packages dependent on `@prismicio/richtext` should be upgraded to `@prismicio/client/richtext`.

For more details behind this PR's motivation, see this thread: https://github.com/prismicio/prismic-richtext/pull/63#discussion_r1296975156

### Migration guide

1. Update to the latest version of `@prismicio/client`:

   ```sh
   npm install @prismicio/client@latest
   ```

1. Replace all `@prismicio/richtext` imports with `@prismicio/client/richtext`:

   ```diff
   - import { serialize } from "@prismicio/richtext";
   + import { serialize } from "@prismicio/client/richtext";
   ```

1. Uninstall `@prismicio/richtext`:

   ```sh
   npm uninstall @prismicio/richtext
   ```

## Checklist:

- [x] My change requires an update to the official documentation.
- [x] All [TSDoc](https://tsdoc.org) comments are up-to-date and new ones have been added where necessary.
- [x] All new and existing tests are passing.

🦕

PR #326, opened 2023-09-29
feat: support extended media API response

lihbr:
## Types of changes

## Description

Context: https://prismic-team.slack.com/archives/C014VAACCQL/p1695984909427609

I considered the new properties as optional because users will need to trigger a master ref release for them to appear, but maybe we don't need to? 🤔

This is to be released in sync with API changes in November.

## Checklist:

### DIFF
--- a/src/types/value/image.ts
+++ b/src/types/value/image.ts
@@ -12,18 +12,27 @@ export type ImageFieldImage<State extends FieldState = FieldState> =
 	State extends "empty" ? EmptyImageFieldImage : FilledImageFieldImage;
 
 export interface FilledImageFieldImage {
+	id?: string;
 	url: string;
 	dimensions: {
 		width: number;
 		height: number;
+	};
+	edit?: {
+		x: number;
+		y: number;
+		zoom: number;
+		background: string;
 	};
 	alt: string | null;
 	copyright: string | null;
 }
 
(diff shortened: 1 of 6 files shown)


angeloashmore:
I think we can remove the `?` from the properties in this case. Developers are unlikely to use the properties anytime soon, if ever, at which point their master ref has probably already been updated.

Following that logic, should we also remove the `?` from `Slice`'s `id` property?

Otherwise, everything looks good to merge when the API is updated.

lihbr:
Should be good @angeloashmore, type tests are failing however due to https://github.com/prismicio/prismic-client/pull/318, do you want me to finish it?

angeloashmore:
@lihbr Awesome! I made a few changes to fix the tests, which uncovered a small bug.

- I added `// @ts-expect-error` to some of the tests that use `@prismicio/mock`. Because `mock` uses `@prismicio/client` as a dependency, it doesn't have the updated types that `@prismicio/client` is looking for. We can remove the `// @ts-expect-error` comments once we publish these type changes.

- There was a bug in how `asImageWidthSrcSet()` extracted thumbnails. It tried to use the `id` and `edit` properties as thumbnails, which threw an error, of course.

With those fixes, we are good to merge once the updates go live!

PR #327, opened 2023-10-02
feat: support expired and not-found ref API errors

angeloashmore:
## Types of changes

## Description

This PR adds support for two new ref-related errors that may be returned by Rest API V2:

- **`RefNotFoundError`**: Thrown when a query's ref does not exist.
- **`RefExpiredError`**: Thrown when a query's ref is expired (generally older than 5 minutes if it is not the master ref or a release).

This is a non-breaking change since the errors were previously thrown as `ForbiddenError`s. The new error types extend from `ForbiddenError` and will continue to work with `instanceof ForbiddenError`.

The new errors will extend from `PrismicError` in a future major version for better semantics.

## Checklist:

🫎


src/errors/RefNotFoundError.ts
```diff
@@ -0,0 +1,8 @@
+import { PrismicError } from "./PrismicError";
+
+type RefNotFoundErrorAPIResponse = {
+	type: "api_notfound_error";
+	message: string;
+};
+
+export class RefNotFoundError extends PrismicError<RefNotFoundErrorAPIResponse> {}
```
lihbr, line 8: Can we make this extend `NotFoundError` perhaps? This makes it a non-breaking change and is still semantically correct(?)

lihbr (approved):
Just a small suggestion so we can release that as a minor, but look good to me otherwise :)

src/errors/RefNotFoundError.ts
angeloashmore: The API was returning 403 responses before the recent change, which means a `ForbiddenError` was thrown instead. We could still extend from `ForbiddenError` instead to prevent the breaking change.

What do you think? I'm okay with extending `ForbiddenError` provide we add a TODO to change it to `PrismicError` in v8. In my opinion, `NotFoundError` should be completely separate from `RefNotFoundError` since the API's payloads may be different.

angeloashmore: I updated the PR ([see this commit](https://github.com/prismicio/prismic-client/pull/327/commits/cdbedd79bfa31e0e5f5aff249f1e5ca0403af57e)) and updated the PR body. I can still rollback or change the implementation if you have any other thoughts. 🙂

lihbr: OK yes, perfect!

Issue #345, opened 2024-07-24
Use of `rehype-parse` causes CommonJS builds to fail


strmer15:
### Versions

- @prismicio/client: 7.7.4
- node: 20.15.1

### Reproduction

Get the latest version of `@prismicio/client` and use it in a CommonJS build with Webpack.

<details open>
<summary>Additional Details</summary>
<br>

</details>

### Steps to reproduce

Use `@prismicio/client` 7.7.x in a CommonJS build with Webpack.

### What is expected?

I can use `@prismicio/client` in a CommonJS Webpack build without any errors.

### What is actually happening?

It's throwing `ERR_REQUIRE_ESM` when trying to load `rehype-parse`. The `rehype-parse` package is ESM only, so by including it this package has essentially become ESM only.

```
/home/cplummer/work/beethoven/node_modules/.pnpm/@prismicio+client@7.7.4/node_modules/@prismicio/client/dist/helpers/unstable_htmlAsRichText.cjs:121
undefined
      ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /home/cplummer/work/beethoven/node_modules/.pnpm/rehype-parse@9.0.0/node_modules/rehype-parse/index.js from /home/cplummer/work/beethoven/node_modules/.pnpm/@prismicio+client@7.7.4/node_modules/@prismicio/client/dist/helpers/unstable_htmlAsRichText.cjs not supported.
Instead change the require of index.js in /home/cplummer/work/beethoven/node_modules/.pnpm/@prismicio+client@7.7.4/node_modules/@prismicio/client/dist/helpers/unstable_htmlAsRichText.cjs to a dynamic import() which is available in all CommonJS modules.
    at Object.newLoader [as .js] (/home/cplummer/work/beethoven/node_modules/.pnpm/pirates@4.0.6/node_modules/pirates/lib/index.js:121:7)
    at Module.<anonymous> (/home/cplummer/work/beethoven/node_modules/.pnpm/@prismicio+client@7.7.4/node_modules/@prismicio/client/dist/helpers/unstable_htmlAsRichText.cjs:3:21)
    at Object.newLoader [as .js] (/home/cplummer/work/beethoven/node_modules/.pnpm/pirates@4.0.6/node_modules/pirates/lib/index.js:121:7) {
  code: 'ERR_REQUIRE_ESM'
}

Node.js v20.15.1
```

lihbr:
[...]

PR #346, opened 2024-07-26
fix: preserve CommonJS compatibility

lihbr:
**Resolves**: #345

This is another option to #347. Only one of the two PR should be merged.

### Description

This PR brings back CommonJS compatibility for `@prismicio/client`. Unified dependencies introduced in v7.7.0 are ESM-only and need to be compiled down to CommonJS by `@prismicio/client`. This is achieved by moving the dependencies to the `devDependencies` map, saving package users from installing them for no purpose since they are now inlined in the `dist` folder.

For the above to work, this PR also has to do the following:
- Rollup needs to be told to explicitly use the Node-compatible export of `decode-named-character-reference` (the Node-compatible export being browser-compatible too). Rollup would default to using the its browser dedicated export otherwise which is incompatible.
- `css-selector-parser` and `parse5` need to remain external and therefore are declared as regular `dependencies`. This is done because otherwise when attempting to inline them in `dist`, Rollup/ESBuild would replace their type imports (e.g. https://github.com/mdevils/css-selector-parser/blob/master/src/render.ts#L1) with bare imports, potentially causing bundlers bundling `@prismicio/client` later on to worry why it's declared side-effect free while have bare imports.
   **Note:** I find this one quite weird to be honest. I tried playing with TypeScript compiler options `importsNotUsedAsValues` and `verbatimModuleSyntax` with no luck. I don't know why it keeps on creating bare imports, those packages' `dist` don't even have them :(

### Checklist

### Preview

```
npm install @prismicio/client@7.7.5-alpha.0
```

### How to QA [^1]

1. Checkout and build this branch:
   ```bash
   git checkout lh/cjs-compat
   npm run build
   ```
2. At the root of `@prismicio/client`:
   1. Create an `index.cjs` file with the following content and run it with `node index.cjs`
      ```javascript
      const prismic = require("./dist/index.cjs")
      ```
   3. Create an `index.js` file with the following content and run it with `node index.js`
      ```javascript
      import * as prismic from "./dist/index.js"
      ```

[^1]:


lihbr:
Closed in favor of #347

PR #347, opened 2024-07-29
feat: remove `*AsRichText` helpers

lihbr:
**Resolves**: #345

This is another option to #346. Only one of the two PR should be merged.

### Description

This PR removes `unstable_htmlAsRichText` and `unstable_markdownAsRichText` helpers from `@prismicio/client`.

These migration-related helpers will be made available through a dedicated package to limit their impact on this core Prismic SDK: https://github.com/prismicio/prismic-migrate (to be made public)

This is **not** considered a breaking change because the helpers were considered unstable and not following SemVer. In the event you were (already) using them, migrate to `@prismicio/migrate`.

### Checklist

### How to QA [^1]

Double-check for any remaining of the previous helpers. Some fixes on types and alike were made when adding them so I didn't simply revert to a previous commit (and we should own our mistakes)

[^1]:


angeloashmore (approved):
LGTM!

PR #349, opened 2024-08-23
feat: add client internal query params

lihbr:
**Resolves**: N/A

### Description

This pull request adds support for client internal query parameters for the API to use. For more context, see: https://prismic-team.slack.com/archives/C030DCL4JNR/p1724406092961109

- `x-d` is forwarded when `process.env.NODE_ENV === "development`
- `x-c` is forwarded with the attached client version

### Checklist

### Preview

N/A

### How to QA [^1]

[^1]:

### DIFF
--- a/src/buildQueryURL.ts
+++ b/src/buildQueryURL.ts
@@ -1,16 +1,18 @@
 import { castArray } from "./lib/castArray"
 import { devMsg } from "./lib/devMsg"
 
+import { version } from "../package.json"
+
 /**
  * The query parameter used to indicate if the client is in development mode to
  * the API.
  */
-export const PRISMIC_DEV_PARAM = "x-d"
+const PRISMIC_DEV_PARAM = "x-d"
 
 /**
  * The query parameter used to indicate the version of the client to the API.
  */
-export const PRISMIC_CLIENT_VERSION_PARAM = "x-c"
+const PRISMIC_CLIENT_VERSION_PARAM = "x-c"
 
 /**
  * Create a union of the given object's values, and optionally specify which
@@ -203,20 +205,6 @@ export interface QueryParams {
 	 * {@link https://prismic.io/docs/route-resolver}
 	 */
 	brokenRoute?: string
-
-	/**
-	 * Whether or not the client is running in a development environment.
-	 *
-	 * @internal
-	 */
-	[PRISMIC_DEV_PARAM]?: number
-
-	/**
-	 * The client version used to make the request.
-	 *
-	 * @internal
-	 */
-	[PRISMIC_CLIENT_VERSION_PARAM]?: string
 }
 
 /**
@@ -397,6 +385,12 @@ export const buildQueryURL = (
 				castArray<string | number | Route | Ordering>(value).join(","),
 			)
 		}
+	}
+
+	url.searchParams.set(PRISMIC_CLIENT_VERSION_PARAM, `js-${version}`)
+
+	if (process.env.NODE_ENV === "development") {
+		url.searchParams.set(PRISMIC_DEV_PARAM, "1")
 	}
 
 	return url.toString()
--- a/src/buildQueryURL.ts
+++ b/src/buildQueryURL.ts
@@ -2,6 +2,17 @@ import { castArray } from "./lib/castArray"
 import { devMsg } from "./lib/devMsg"
 
 /**
+ * The query parameter used to indicate if the client is in development mode to
+ * the API.
+ */
+export const PRISMIC_DEV_PARAM = "x-d"
+
+/**
+ * The query parameter used to indicate the version of the client to the API.
+ */
+export const PRISMIC_CLIENT_VERSION_PARAM = "x-c"
+
+/**
  * Create a union of the given object's values, and optionally specify which
  * keys to get the values from.
  *
@@ -192,6 +203,20 @@ export interface QueryParams {
 	 * {@link https://prismic.io/docs/route-resolver}
 	 */
 	brokenRoute?: string
+
+	/**
+	 * Whether or not the client is running in a development environment.
+	 *
+	 * @internal
+	 */
+	[PRISMIC_DEV_PARAM]?: number
+
+	/**
+	 * The client version used to make the request.
+	 *
+	 * @internal
+	 */
+	[PRISMIC_CLIENT_VERSION_PARAM]?: string
 }
 
 /**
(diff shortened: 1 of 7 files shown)


lihbr:
Updated as per discussion 👌

src/buildQueryURL.ts
```diff
@@ -1,6 +1,17 @@
...
+/**
+ * The query parameter used to indicate if the client is in development mode to
+ * the API.
+ */
+export const PRISMIC_DEV_PARAM = "x-d"
+
+/**
+ * The query parameter used to indicate the version of the client to the API.
+ */
+export const PRISMIC_CLIENT_VERSION_PARAM = "x-c"
```
angeloashmore, line 13: ❓ #ask: Should we make this strategy compatible with other languages? With the current version, analytics around `x-c` assume `@prismicio/client` is being used.

What happens if we want to track usage of other clients, like the PHP client? We only officially support the JavaScript ecosystem, but we may want to track usage of the languages to better understand how often other languages are used.

We may want to consider a URL scheme that works across languages.

See Imgix's `ix-lib` parameter for an example: https://github.com/imgix/js-core/blob/bfca36f5cac1833ff2ff390de4e42a15b215a89c/src/index.js#L114

src/buildQueryURL.ts
```diff
@@ -192,6 +203,20 @@ export interface QueryParams {
...
+	 * @internal
+	 */
+	[PRISMIC_DEV_PARAM]?: number
+
+	/**
+	 * The client version used to make the request.
+	 *
+	 * @internal
+	 */
+	[PRISMIC_CLIENT_VERSION_PARAM]?: string
```
angeloashmore, line 219: 💡 #idea: Rather than accept these as arguments, it might make more sense for `buildQueryURL()` to inject these automatically. WDYT?

src/createClient.ts
```diff
@@ -460,6 +466,22 @@ export class Client<TDocuments extends PrismicDocument = PrismicDocument> {
...
+		"ref" | "integrationFieldsRef" | "accessToken" | "routes"
+	> =
+		process.env.NODE_ENV === "development"
+			? {
+					[PRISMIC_DEV_PARAM]: 1,
+					[PRISMIC_CLIENT_VERSION_PARAM]: version,
+				}
+			: {
+					[PRISMIC_CLIENT_VERSION_PARAM]: version,
+				}
```
angeloashmore, line 483: 💡 #idea: This seems unnecessary as an instance property. Couldn't it be a constant? Or even better, passed directly to `buildQueryURL()` as arguments? It's simpler and requires less code.

```ts
buildQueryURL(this.endpoint, {
  "x-d": process.env.NODE_ENV === "development" ? 1 : undefined,
  "x-c": version
})
```

Disregard this if we decide to inject the parameters within `buildQueryURL()` directly.

test/client-buildQueryURL.test.ts
```diff
@@ -23,6 +23,8 @@ it("builds a query URL using the master ref", async (ctx) => {
 		ref,
 	})
 	url.searchParams.delete("integrationFieldsRef")
+	url.searchParams.delete("x-c")
+	url.searchParams.delete("x-d")
```
angeloashmore, line 27: ❓ #ask: I don't remember why we delete `integrationFieldsRef`. Shouldn't we check for all expected params, including `x-c` and `x-d`?

src/buildQueryURL.ts
lihbr: OK, what about `js@vX.Y.Z` as a syntax?

src/createClient.ts
lihbr: Let's inject it directly and keep the client out of that

test/client-buildQueryURL.test.ts
lihbr: I think it's because each tests gets a random one assigned by the mocked API for `integrationFieldsRef`. I'll rewrite it to expect `x-c` and `x-d`

lihbr: Done!

src/buildQueryURL.ts
lihbr: Agreed, refactored!

lihbr:
Refactored as per discussions! (thanks, it's much leaner like this 🙏)

Let me know about it and I'm happy to ship afterward :)

test/buildQueryURL.test.ts
```diff
@@ -247,10 +252,28 @@ it("supports array routes param", () => {
...
 	)
 })
 
+it("forwards `x-c` header in production", () => {
+	expect(prismic.buildQueryURL(endpoint, { ref: "ref" })).toBe(
+		`https://qwerty.cdn.prismic.io/api/v2/documents/search?ref=ref${xClientVersionParam}`,
+	)
+})
+
+it("forwards `x-c` and `x-d` headers in development", () => {
```
angeloashmore, line 265: ⚠️ #issue: Tiny fix:

```suggestion
it("forwards `x-c` param in production", () => {
	expect(prismic.buildQueryURL(endpoint, { ref: "ref" })).toBe(
		`https://qwerty.cdn.prismic.io/api/v2/documents/search?ref=ref${xClientVersionParam}`,
	)
})

it("forwards `x-c` and `x-d` param in development", () => {
```

test/client-buildQueryURL.test.ts
```diff
@@ -31,6 +36,39 @@ it("builds a query URL using the master ref", async (ctx) => {
...
+	url.searchParams.delete("integrationFieldsRef")
+	url.searchParams.sort()
+	expectedSearchParams.sort()
+
+	expect(url.host).toBe(new URL(client.endpoint).host)
+	expect(url.pathname).toBe("/api/v2/documents/search")
+	expect(url.searchParams.toString()).toBe(expectedSearchParams.toString())
+
+	process.env = originalEnv
+})
```
angeloashmore, line 70: 💡 #idea: We could focus this test on just the `x-d` parameter since the rest of the test is checked in other tests.

```suggestion
it("includes the `x-d` param in development", async (ctx) => {
	const originalEnv = { ...process.env }
	process.env.NODE_ENV = "development"

	mockPrismicRestAPIV2({ ctx })

	const client = createTestClient()
	const res = await client.buildQueryURL()
	const url = new URL(res)

	expect(url.searchParams.get("x-d")).toBe("1")

	process.env = originalEnv
})
```

angeloashmore (approved):
Looks good to me! I left two small suggestions, but feel free to merge with or without them. 🙂

lihbr:
Thanks, let's release that!

Issue #333, opened 2024-03-19
Automatically handle expired refs


angeloashmore:
### Is your feature request related to a problem? Please describe.

`@prismicio/client` throws a `RefExpiredError` when a request's ref is expired. The client typically fetches the active ref automatically before sending a content request, but some environments cache the ref. For example, Next.js caches `fetch()` requests, which could result in a request made with a stale ref.

Because `@prismicio/client` handles fetching the ref automatically, developers are not expecting the ref to be expired. Instead, they expect the client to fetch the content they requested.

### Describe the solution you'd like

Currently, the error provides the master ref as part of the message:

```
The provided ref is no longer accessible, it has expired. Master ref is: xxxxxxxxxxxxxx
```

We can extract the master ref from this error message and retry the request with the new ref.

When the client retries a request, the developer should be notified through a console warning.

If the retried request *also* has an expired ref, we can continue retrying until a valid ref is found. We can limit the number of retries to prevent infinite loops.

When a developer explicitly provides a ref, such as through `defaultParams`, a preview session, or on a per-query basis, the ref should be used as provided. The retry logic and console message should be skipped. This avoids a case where a developer is sure they want to use a specific ref. We should respect that decision and let them handle the error or change their code.

### Describe alternatives you've considered

Developers can already catch `RefExpiredError` and handle the retry themselves. That is a solution to a leaky abstraction and is not something we should ask developers to do.

Developers can also configure their environments to not cache requests to `/api/v2`. This kind of configuration can be tricky to set up, however, and may introduce additional headaches.

### Additional context

This error most often comes from Next.js websites using the App Router. The App Router caches `fetch()` requests when configured properly, and [we recommend caching all Prismic requests](https://prismic.io/docs/fetch-data-nextjs#caching-with-the-app-router).

WYGIN:
@angeloashmore I am interested in contributing to this feature, would you mind let me know if this feature has any high priority, as it would take sometime for me to fix this issue.
Thank You.

angeloashmore:
Completed in #356.

PR #356, opened 2024-10-29
feat: retry invalid/expired refs

angeloashmore:
**Resolves**: #333

### Description

This PR adds a retry mechanism when a `RefNotFoundError` or `RefExpiredError` is thrown in a query method. The client will retry a maximum of 3 times before throwing.

A warning is logged when retrying:

```ts
"The ref (xxx) was expired. Now retrying with the latest master ref (yyy). If you were previewing content, the response will not include draft content."
```

This PR also clears the `/api/v2` value cached for 5s when we know a stale value is saved. This prevents a small edge case where calls within 5s would have all failed if the ref was invalid or expired.

### Checklist

### Preview

There are no changes to existing code. Queries appear as before.

### How to QA [^1]

See the tests.

[^1]:

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -74,6 +74,13 @@ export const GET_ALL_QUERY_DELAY = 500
 const DEFUALT_RETRY_AFTER_MS = 1000
 
 /**
+ * The maximum number of attemps to retry a query with an invalid ref before
+ * halting. We allow multiple attemps as each attemp may return a different ref.
+ * Capping the number of attemps prevents infinite loops.
+ */
+const MAX_INVALID_REF_RETRY_ATTEMPS = 3
+
+/**
  * Extracts one or more Prismic document types that match a given Prismic
  * document type. If no matches are found, no extraction is performed and the
  * union of all provided Prismic document types are returned.
@@ -1696,6 +1703,7 @@ export class Client<
 	 */
 	private async _get<TDocument extends TDocuments>(
 		params?: Partial<BuildQueryURLArgs> & FetchParams,
+		attemptCount = 0,
 	): Promise<{ data: Query<TDocument>; url: string }> {
 		const url = await this.buildQueryURL(params)
 
@@ -1705,7 +1713,10 @@ export class Client<
 			return { data, url }
 		} catch (error) {
 			if (
-				!(error instanceof RefNotFoundError || error instanceof RefExpiredError)
+				!(
+					error instanceof RefNotFoundError || error instanceof RefExpiredError
+				) ||
+				attemptCount >= MAX_INVALID_REF_RETRY_ATTEMPS - 1
 			) {
 				throw error
 			}
@@ -1722,7 +1733,7 @@ export class Client<
 				`The ref (${badRef}) was ${issue}. Now retrying with the latest master ref (${masterRef}). If you were previewing content, the response will not include draft content.`,
 			)
 
-			return await this._get({ ...params, ref: masterRef })
+			return await this._get({ ...params, ref: masterRef }, attemptCount + 1)
 		}
 	}
 
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -262,6 +262,12 @@ export class Client<
 	#repositoryName: string | undefined
 
 	/**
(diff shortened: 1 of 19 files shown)


src/Client.ts
```diff
@@ -531,7 +537,36 @@ export class Client<
...
+
+			if (this.#retriedRefs.has(latestRef)) {
+				// We only allow retrying a ref once. We'll
+				// likely get the same response on future
+				// retries and don't want to get stuck in a
+				// loop.
+				throw error
+			}
+
+			this.#retriedRefs.add(latestRef)
```
lihbr, line 561: I'm afraid this could create some concurrency issues when we dedupe requests:

- T0
  - Page Foo fetches nav document
  - Page Bar fetches nav document
- T1
  - Requests get deduped (only one is made) and fail at the same time with the same error
- T2
  - Page Foo retries the request with the latest ref
  - Page Bar cannot retries the request with the latest ref because it's already being retried by Foo
  - Request cannot be deduped

...or when multiple requests (e.g. `Promise.all(...)`) fail at the same time, which I believe is symptomatic of this kind of error(?)

Maybe we should base this off whether of not a `ref` parameter was provided to the `get` function? i.e. no `ref`: retries with an explicit ref, `ref`: cannot retry

angeloashmore: You're right; it would likely cause some issues when queries are deduplicated. I removed the `retriedRefs` array and check and simply trust that the provided refs won't cause a loop.

Here's the comment I sent to you [in Slack](https://prismic-team.slack.com/archives/C014VAACCQL/p1730251254283029):

> I removed the logic that prevents the same ref from being retried more than once. You are right about concurrency issues since the retry logic is in the query method, not in the deduplicated fetch job. I'd like to keep the retry in the method so it is close to its intended usage.
>
>I'm wondering if we should still include some kind of loop prevention. I refactored the implementation you reviewed yesterday such that we now have a private `_get()` method that preforms the retry. Since we have that now, it's easy for us to track how many retries and halt if we get too deep.
>
>I think we should include some kind of loop protection to be safe, but I wanted to get your thoughts first.

---

> Maybe we should just retry directly instead of recursively calling the function after that?

I'd rather not retry directly since it means we can only follow the API's response once. If a website has a very stale cache with multiple chained invalid ref responses, we should be able to use each of them.

We log a warning each time we retry, so devs will still be nudged to fix their caching issue.

---

> Or base the retry off whether or not we had an explicit `ref` parameter(?)

A retry may be necessary whether or not an explicit `ref` is provided. Websites may have a stale `/api/v2`, which would result in a stale master ref being used implicitly.

---

> 💡 #idea: Maybe we should clear the repository cache when this happens so that other requests (during the cache's 5s lifespan) can fetch a fresh ref?

Maybe! This fix is really intended for Next.js' indefinite cache, which works outside the client's internal 5s cache.

I can't think of any downside to clearing the internal cache. It would resolve the tiny chance that a burst of requests from the same client results uses a stale, internally cached ref. If you can't think of any downsides, I'll add it in.

lihbr: Thanks for the clarification!

- I agree we should have some kind of loop prevention yes ♻️ 
- I'm aligned with clearing the cache, it's mostly free and will clear some edge cases 👌

lihbr (approved):
LGTM!

Just some kind of loop prevention as discussed would be great, maybe through a `triedRefs` argument on `_get()`

And repository cache purging also ☺️

angeloashmore:
@lihbr I made the following changes:

- Retries are limited to 3 attempts.
- The cached repository value is cleared when we know a stale `/api/v2` was used.
- Added tests for those cases.

You can see the changes here: https://github.com/prismicio/prismic-client/pull/356/files/d10ebe6d8fe3f5d5461e3e565d8d4c93370e60ba..f3ae551cd9da9dee493971db58c8c4f003f8ee87?diff=split&w=1

raRaRa:
It would be AWESOME if this could also bypass the cache in the retry if the ref is expired. It has caused so much pain in my NextJS projects where websites suddenly start to "crash" due to this issue. 

I was deploying a new version of a site and all of the pages started crashing cause Vercel stores cached data between deployments. I had to delete all the cache manually.

PR #364, opened 2024-12-11
Link variant

dani-mp:
**Resolves**: 
Part of https://linear.app/prismic/issue/DT-2469/aauser-when-i-create-a-link-field-and-allow-variants-i-can-see-an.

### Checklist

### How to QA [^1]

[^1]:


src/types/value/link.ts
```diff
@@ -25,6 +25,7 @@ export type EmptyLinkField<
 > = {
 	link_type: Type | string
 	text?: string
+	variant?: string
```
angeloashmore, line 28: ❓ #ask: Does this imply someone can choose a link variant without providing a URL, similar to providing text without a URL? Just asking for my own knowledge; the code looks good.

src/types/value/contentRelationship.ts
```diff
@@ -45,4 +45,5 @@ export interface FilledContentRelationshipField<
 	isBroken?: boolean
 	data?: DataInterface
 	text?: string
+	variant?: string
```
angeloashmore, line 48: ⚠️ #issue: From [this Slack message](https://prismic-team.slack.com/archives/C02L3FN3AJK/p1733937871356369?thread_ts=1733933751.211369&cid=C02L3FN3AJK), it seems we are only supporting `variant` on link fields, not content relationship and link to media fields.

src/types/value/linkToMedia.ts
```diff
@@ -25,4 +25,5 @@ export interface FilledLinkToMediaField {
 	height?: string | null
 	width?: string | null
 	text?: string
+	variant?: string
```
angeloashmore, line 28: ⚠️ #issue: From [this Slack message](https://prismic-team.slack.com/archives/C02L3FN3AJK/p1733937871356369?thread_ts=1733933751.211369&cid=C02L3FN3AJK), it seems we are only supporting `variant` on link fields, not content relationship and link to media fields.

src/types/value/contentRelationship.ts
angeloashmore: I wonder if we need to add the `variant` property in `src/types/value/link.ts` as an intersection. Something like this pseudo-code:

```ts
type LinkField = (WebLinkField | LinkToMediaField | ContentRelationshipField) & {
  variant?: string
}
```

That keeps `variant` out of link to media and content relationship fields.

angeloashmore (commented):
Looks good so far! I left a few questions and suggestions.

src/types/value/link.ts
dani-mp: yes, same rationale

src/types/value/contentRelationship.ts
dani-mp: first message: i think what we want is to not add this option to the link to media or content relationship concrete fields, but you can still create a _generic_ link without the select and then in page builder select one of the types

second message: i understood from a conversation with @xrutayisire the other day that one might want to use a concrete version of the field and it should already contain all properties, as they're all exported individually

PR #366, opened 2024-12-13
fix: fix link types value + model

xrutayisire:
### Description

This PR objective is to clean up and fix the link types (model + value).

This PR initially assumes that we will support `allowText` and `variants` for Link To Media and Generic Link.
We are waiting for a final product decision and, of course, this will impact the code if we decide to drop support, as it's the case for Content Relationship.

**Model**

- `allowText` should not be defined on Content Relationship
- `variants` should be defined for Link To Media.

**Value**

- Ensure all `Filled` links have text and variants as they are re-used everywhere to define what a link even coming from a generic link can be.
- Ensure `EmptyLinkField` doesn't by default have `text` and `variant` as it's used in `ContentRelationshipField`
- Omit `text` and `variant` manually when types are used for Rich Text

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/src/types/value/contentRelationship.ts
+++ b/src/types/value/contentRelationship.ts
@@ -1,7 +1,7 @@
 import type { AnyRegularField, FieldState } from "./types"
 
 import type { GroupField } from "./group"
-import type { EmptyLinkField, LinkType } from "./link"
+import type { EmptyLinkField, LinkType, WithLinkAdditionalData } from "./link"
 import type { SliceZone } from "./sliceZone"
 
 /**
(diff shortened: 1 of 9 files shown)


src/types/value/contentRelationship.ts
```diff
@@ -22,18 +22,21 @@ export type ContentRelationshipField<
...
  */
-export interface FilledContentRelationshipField<
+export type FilledContentRelationshipField<
 	TypeEnum = string,
 	LangEnum = string,
 	DataInterface extends
 		| Record<string, AnyRegularField | GroupField | SliceZone>
 		| unknown = unknown,
-> {
+> = WithLinkAdditionalData<{
```
dani-mp, line 39: not sure i understand this one, we wrap FilledContentRelationshipField with WithLinkAdditionalData, adding those two additional props to it? i thought we didn't want this

src/types/value/link.ts
```diff
@@ -15,6 +15,16 @@ export const LinkType = {
 	Web: "Web",
 } as const
 
+/**
+ * Additional optional data for a link field.
+ *
+ * @typeParam Link - Link field.
+ */
+export type WithLinkAdditionalData<Link> = Link & {
```
dani-mp, line 23: what about
```suggestion
export type WithAdditionalLinkAttributes<Link> = Link & {
```
another option: WithAdditionalLinkProps

dani-mp (commented):
left a question and a suggestion

src/types/value/contentRelationship.ts
xrutayisire: Yes, we want this because it's used everywhere internally to express what is a valid Link To Document even for Generic Link.

dani-mp: sorry, i don't understand

for link the generic link, we wrap ContentRelationshipField with the additional data, so it's covered when we have a link to document via generic link

but why do we want this as well when it's a standalone content relationship field? when devs use this one it'll have the optional properties even though the model doesn't allow them

dani-mp: i see now

filled contet rel. has the add. props
we remove them from content rel.
we add them again when we use content rel. inside the generic link

Issue #369, opened 2024-12-21
Prismic Client ignores and overwrites 'page' query param for getAll methods


malee31:
### Versions

- @prismicio/client: v7.13.1
- node: v20.11.0

### Reproduction
Create 4 documents of the same type, configure the following snippet, and run.

```js
import * as prismic from "@prismicio/client";

// Replace these with your Prismic repository name and a valid custom page type
const repo = "my-repo";
const documentType = "pages";

const client = prismic.createClient(repo, {
  fetch: (...args) => {
    // For inspecting the endpoint being hit
    if(args[0].includes(documentType)) {
      console.log(`GET: ${args[0]}`);
    }

    return fetch(...args);
  },
});

// IIFE attempting to fetch the first two pages and print their IDs
(async () => {
  const pageOne = await client.getAllByType(documentType, {
    page: 1,
    limit: 2,
    pageSize: 2,
    orderings: {
      field: "document.first_publication_date",
      direction: "desc",
    },
  });

  const pageTwo = await client.getAllByType(documentType, {
    page: 1,
    limit: 2,
    pageSize: 2,
    orderings: {
      field: "document.first_publication_date",
      direction: "desc",
    },
  });

  // None of these IDs should not be the same if paging is working properly
  // Currently, they both print the same results
  console.log(`Page One: [${pageOne.map(res => res.id).join(", ")}]`);
  console.log(`Page Two: [${pageTwo.map(res => res.id).join(", ")}]`);
})();
```

### Steps to reproduce
1. Log into or create a new Prismic Repository
2. Create a new Page Type if one does not already exist
3. Publish at least 4 documents using the aforementioned page type
4. Copy the code above into a new folder as `index.js`
5. Initialize a Node Project repository with `npm init`
6. Install the Prismic Client with `npm install @prismicio/client`
7. Change the constants `repo` and `documentType`
8. Run `node index.js`

### What is expected?
From the documentation [for query method params](https://prismic.io/docs/technical-reference/prismicio-client/v7#query-methods-params), we are able to pass the `?page=#` to the REST API to paginate our results with `pageSize`.

So passing in `page: 2` should return the second page of results.

### What is actually happening?
Regardless of which page is being queried for through the client `getAll*` functions, results start from the very first page.
[...]

PR #370, opened 2024-12-21
fix: allow "page" param to properly pass through getall methods

malee31:
**Resolves**: #369

### Description

The `dangerouslyGetAll` method (and all `getAll*` methods utilizing it) now uses the `page` param passed in as the starting point for its search results instead of only searching from the first page.
 
This fixes the offset pagination for the client without having to directly create and send out API requests with `fetch()`.

### Checklist

No existing tests fail as a result of this change.

- [x] A comprehensive Linear ticket, providing sufficient context and details to facilitate the review of the PR, is linked to the PR.

### How to QA [^1]

Utilize the snippet of code provided in the original issue on the new client.

Changes have not yet been tested on a real Prismic project. The code was read and traced through line by line.

[^1]:

### DIFF
--- a/src/Client.ts
+++ b/src/Client.ts
@@ -605,6 +605,7 @@ export class Client<
 	): Promise<TDocument[]> {
 		const { limit = Infinity, ...actualParams } = params
 		const resolvedParams = {
+			page: undefined,
 			...actualParams,
 			pageSize: Math.min(
 				limit,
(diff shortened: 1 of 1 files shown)


angeloashmore:
Hi @malee31, thanks for the PR. The `getAll*` methods are designed to fetch all documents for a particular query. We provide a way to limit the total number of documents returned via the `limit` parameter. However, we feel configuring a starting offset via the `page` parameter is too specific of a use case to introduce into the client.

If you are performing pagination, you are best off using the non-`getAll*` methods since they provide pagination metadata.

Please let us know if you have any questions and we'll try to help. Thanks!

PR #373, opened 2025-01-14
fix: properly update document's title with `updateDocument`

dani-mp:
**Resolves**: 

### Description

While using the migration client to write integration tests for the Migration API, I noticed that this test was not passing (last line):
```ts
it("Can update a document with a permanent token", async () => {
  // Verify if the last test was successful
  expect(document).toBeDefined()

  const newTitle = `Document updated`

  const migration = prismicClient.createMigration()
  const migrationDocument = migration.updateDocument(
    {
      ...document,
      data: { text: "lorem ipsum" },
    },
    newTitle,
  )
  await writeClient.migrate(migration)

  const documentResponse = await coreApiClient.get<CoreApiDocument>(
    `/core/documents/${migrationDocument.document.id}`,
  )
  expect(documentResponse).toHaveStatusCode(200)
  expect(documentResponse.data?.title).toEqual(newTitle) // 💥
})
```
As you can see, the migration updates the document, but its title has never been updated. This PR should fix it.

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/src/WriteClient.ts
+++ b/src/WriteClient.ts
@@ -504,6 +504,7 @@ export class WriteClient<
 				// in case the document already existed during the previous step.
 				{
 					...doc.document,
+					documentTitle: doc.title,
 					data: await resolveMigrationDocumentData(
 						doc.document.data,
 						migration,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [7.15.0-alpha.0](https://github.com/prismicio/prismic-client/compare/v7.14.0...v7.15.0-alpha.0) (2025-01-14)
+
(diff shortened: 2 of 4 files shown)


src/WriteClient.ts
```diff
@@ -504,6 +504,7 @@ export class WriteClient<
 				// in case the document already existed during the previous step.
 				{
 					...doc.document,
+					documentTitle: doc.title,
```
lihbr, line 507: ❓ #ask: Do you think it would make sense to instead update the private `updateDocument` to instead use `document.title`?

I think it could make more sense to do so as `documentTitle` only makes sense in the context of the `createDocument` method arguments 🤔

Happy to update in that direction if you're aligned!

lihbr (commented):
LGTM, just one question!

lihbr (approved):
LGTM, just one question!

src/WriteClient.ts
dani-mp: you mean by passing `doc` directly as a first parameter (which includes the actual document and the title)?

lihbr: Ohhh, ok, no nevermind! I thought we were spreading `doc` but we're spreading `doc.document`, all good

Issue #372, opened 2025-01-14
Link variants should be typable with a generic


lihbr:
### Versions

- @prismicio/client: 7.14.0
- node: 22.0.0

### Steps to reproduce

1. Create a link field with variants with Slice Machine
2. Check out generated TypeScript

### What is expected?

Link variants should be precisely typed when available values are known (always the case for a link field typed after a custom type definition)

This starts by allowing the `LinkField` type to support another generic to type available variants beyond `string` but it's not really ideal without reordering generics today (which would be a breaking change).

From here, `prismic-ts-codegen` could leverage the generic to type variants more precisely the same way it does it for other generic-dependant types (e.g. constrained content relationships)

### What is actually happening?

Generated types

```ts
{
  navigation: prismic.Repeatable<prismic.LinkField>;
}
```

Result
![Image](https://github.com/user-attachments/assets/3d0316bb-c082-454f-a9f2-126acdd1d96d)

Source code
https://github.com/prismicio/prismic-client/blob/c45f28d8c472a0d0a8ecee8d6ac770f52009f0fa/src/types/value/link.ts#L90-L93

angeloashmore:
Completed in #372.

PR #374, opened 2025-01-15
fix: allow generics with link variant

lihbr:
**Resolves**: #372

### Description

Allow link variants to be precisely typed with a generic.

### Checklist

### Preview

See tests: https://github.com/prismicio/prismic-client/pull/374/files#diff-9557d8ecae877dbbb8ecc6dcfbc29d16d7602d12904e3180aed6918bd6eaad8f

### How to QA [^1]

[^1]:

### DIFF
--- a/src/types/value/link.ts
+++ b/src/types/value/link.ts
@@ -23,7 +23,7 @@ export const LinkType = {
  * @typeParam DataInterface - Data fields for the document (filled via the
  *   `fetchLinks` or `graphQuery` query parameter).
  * @typeParam State - State of the field which determines its shape.
- * @typeParam VariantEnum - Variants of the link.
+ * @typeParam Variant - Variants of the link.
  */
 export type LinkField<
 	TypeEnum = string,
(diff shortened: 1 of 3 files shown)


xrutayisire (approved):
Love it! Super nice to have 🙂

src/types/value/link.ts
```diff
@@ -23,6 +23,7 @@ export const LinkType = {
  * @typeParam DataInterface - Data fields for the document (filled via the
  *   `fetchLinks` or `graphQuery` query parameter).
  * @typeParam State - State of the field which determines its shape.
+ * @typeParam VariantEnum - Variants of the link.
```
angeloashmore, line 26: 💡 #idea: Rename `VariantEnum` to `Variant`. We generally don't append the type of data to variable names, like with `State`.

(`DataInterface`, `TypeEnum`, and `LangEnum` should also be cleaned up, but we can leave that out of this PR.)

PR #375, opened 2025-01-21
feat: support table field

levimykel:
**Resolves**: 
https://linear.app/prismic/issue/DT-2584/[m]-aadev-when-i-add-a-table-field-i-can-see-the-according-type

### Description

Adds table field type and value.

### Checklist

### How to QA [^1]

[^1]:


src/helpers/isFilled.ts
```diff
@@ -248,6 +249,17 @@ export const geoPoint = (
...
 }
 
+/**
+ * Determines if a table field is filled.
+ *
+ * @param field - Table field to check.
+ *
+ * @returns `true` if `field` is filled, `false` otherwise.
+ */
+export const table = isNonNullish as (
```
dani-mp, line 259: we might want to do something more sophisticated here, i.e. a table with no head and an empty body is probably "not filled"

src/index.ts
```diff
@@ -227,6 +227,7 @@ export type { NumberField } from "./types/value/number"
 export type { SelectField } from "./types/value/select"
 export type { TimestampField } from "./types/value/timestamp"
 export type { GeoPointField } from "./types/value/geoPoint"
+export type { TableField } from "./types/value/table"
```
levimykel, line 231: I would alphabetize `TableField` and `GeoPointField` here, but it's not a big deal.

src/index.ts
```diff
@@ -288,6 +289,7 @@ export type { CustomTypeModelNumberField } from "./types/model/number"
 export type { CustomTypeModelSelectField } from "./types/model/select"
 export type { CustomTypeModelTimestampField } from "./types/model/timestamp"
 export type { CustomTypeModelGeoPointField } from "./types/model/geoPoint"
+export type { CustomTypeModelTableField } from "./types/model/table"
```
levimykel, line 293: I would alphabetize `CustomTypeModelTableField` and `CustomTypeModelGeoPointField` here, but it's not a big deal.

src/types/value/table.ts
```diff
@@ -0,0 +1,64 @@
...
+import type { RichTextField } from "./richText"
+
+/**
+ * A table field.
+ *
+ * @typeParam State - State of the field which determines its shape.
+ *
+ * @see More details: {@link https://prismic.io/docs/table}
+ */
+export type TableField<State extends FieldState = FieldState> =
```
levimykel, line 12: This content model is out of date. It will need to be updated when you update the `types-internal` alpha to a newer version.

levimykel (commented):
This looks good. I added a couple of comments.

src/helpers/isFilled.ts
xrutayisire: It's true that by default, we will define content for ALL tables, and we will not allow full delete, meaning we will mostly have content for every new pages containing a table. Just with empty cells. Personally, I'm ok to take a table with only empty cell as just a **filled** table field, as even empty cell means table content is defined in my mind. WDYT?

dani-mp: the question is whether as a developer you want to show something on the screen or not if you have a table content defined but with empty rows/cells

for the marketer, having tabl content defined with empty cells means there is content, imo, because there's structure that maybe they want to fill later

but for the consumer side, for the developer, i'm not so sure

xrutayisire: Discussed with Lucie, it's ok not to check every rows / cells to verify the content of the table. But it could be good to add a comment here about the fact table field with rows is considered filled. Even if each cells are an empty text.

lihbr: Fine as is to me yes!

lihbr (commented):
LGTM, thanks for chiming in after the PR merge ☺️

PR #380, opened 2025-03-18
feat: support key for table field content

xrutayisire:
**Resolves**: [DT-2635: \[S\] AADev, I can use the key for each row and cell as a unique identifier](https://linear.app/prismic/issue/DT-2635/[s]-aadev-i-can-use-the-key-for-each-row-and-cell-as-a-unique)

### Description

- We need to have a unique key for each row and cell. This would allow us and user to easily map over rows and cells having a unique identifier and prevent hacky things like `JSON.stringify` or using `index`.

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/src/types/value/table.ts
+++ b/src/types/value/table.ts
@@ -43,7 +43,7 @@ export type TableFieldBody = {
  */
 export type TableFieldHeadRow = {
 	/**
-	 * Unique key of the row (UUID v4 format).
+	 * Unique key of the row.
 	 */
 	key: string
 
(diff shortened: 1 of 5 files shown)


levimykel (approved):
Looks good 🚀

jomifepe (approved):

src/types/value/table.ts
```diff
@@ -42,6 +42,11 @@ export type TableFieldBody = {
  * Represents a row in a table head.
  */
 export type TableFieldHeadRow = {
+	/**
+	 * Unique key of the row (UUID v4 format).
```
angeloashmore, line 46: 💡 #idea: Small nit — I think we can leave out "(UUID v4 format)" in all the TSDocs. The format we use should be irrelevant to users.

xrutayisire: Done 👍

PR #384, opened 2025-05-12
refactor: `isFilled.contentRelationship` function overloads to support strict unions

lihbr:
> [!NOTE]
> This is an implementation of a yet-uncertain approach described in a PRD. We shouldn't move forward with it before we decide to move forward with the strategy described in the PRD.

Resolves: https://www.notion.so/prismic/Sub-PRD-Remove-GraphQuery-Enable-Field-Picking-in-Slice-Machine-UI-1cf3bee08669804a9dd1e6a0050c59a9?pvs=4#1d03bee086698039a908e3a628919242

### Description

This pull request update the `isFilled.contentRelationship` helper's function signature to support a single generic type param allowing for stricter union of content relationship fields to work as expected.

Basically, it allows for this to work:

```typescript
import * as prismic from '@prismicio/client'

type Document = {
  fooOrBar:
    | prismic.ContentRelationshipField<
        'foo',
        string,
        { foo: prismic.RichTextField }
      >
    | prismic.ContentRelationshipField<
        'bar',
        string,
        { bar: prismic.RichTextField }
      >
}

const document: Document = { fooOrBar: { link_type: 'Any' } }

prismic.isFilled.contentRelationship(document.fooOrBar)
```

### Checklist

### Preview

N/A

### How to QA [^1]

Check that updated type tests are sufficient, or try with an updated version of https://www.notion.so/prismic/Sub-PRD-Remove-GraphQuery-Enable-Field-Picking-in-Slice-Machine-UI-1cf3bee08669804a9dd1e6a0050c59a9?pvs=4#1f13bee08669804389addfdef4f12aaf

[^1]:

### DIFF
--- a/src/helpers/documentToLinkField.ts
+++ b/src/helpers/documentToLinkField.ts
@@ -45,9 +45,5 @@ export const documentToLinkField = <
 		...(prismicDocument.data && Object.keys(prismicDocument.data).length > 0
 			? { data: prismicDocument.data }
 			: {}),
-	} as FilledContentRelationshipField<
-		TDocument["type"],
-		TDocument["lang"],
-		TDocument["data"]
-	>
+	}
 }
--- a/src/helpers/documentToLinkField.ts
+++ b/src/helpers/documentToLinkField.ts
@@ -45,5 +45,9 @@ export const documentToLinkField = <
 		...(prismicDocument.data && Object.keys(prismicDocument.data).length > 0
 			? { data: prismicDocument.data }
 			: {}),
-	}
+	} as FilledContentRelationshipField<
+		TDocument["type"],
+		TDocument["lang"],
+		TDocument["data"]
+	>
 }
--- a/src/helpers/isFilled.ts
+++ b/src/helpers/isFilled.ts
@@ -140,31 +140,11 @@ export const linkToMedia = link as (
  *
  * @returns `true` if `field` is filled, `false` otherwise.
  */
-export function contentRelationship<Field extends ContentRelationshipField>(
-	field: Field | null | undefined,
-): field is Exclude<Field, EmptyContentRelationshipField>
-export function contentRelationship<
-	TypeEnum extends string,
-	LangEnum extends string = string,
-	DataInterface extends
-		| Record<string, AnyRegularField | GroupField | SliceZone>
-		| unknown = unknown,
+export const contentRelationship = link as <
+	Field extends ContentRelationshipField,
 >(
-	field:
-		| ContentRelationshipField<TypeEnum, LangEnum, DataInterface>
-		| null
-		| undefined,
-): field is ContentRelationshipField<
-	TypeEnum,
-	LangEnum,
-	DataInterface,
-	"filled"
->
-export function contentRelationship(
-	field: ContentRelationshipField | null | undefined,
-): field is ContentRelationshipField<string, string, unknown, "filled"> {
-	return link(field)
-}
+	field: Field | null | undefined,
+) => field is Exclude<Field, EmptyContentRelationshipField>
 
 /**
  * Determines if a date field is filled.
--- a/src/helpers/isFilled.ts
+++ b/src/helpers/isFilled.ts
@@ -1,5 +1,8 @@
 import type { ColorField } from "../types/value/color"
-import type { ContentRelationshipField } from "../types/value/contentRelationship"
+import type {
+	ContentRelationshipField,
+	EmptyContentRelationshipField,
+} from "../types/value/contentRelationship"
 import type { DateField } from "../types/value/date"
 import type { AnyOEmbed, EmbedField } from "../types/value/embed"
 import type { GeoPointField } from "../types/value/geoPoint"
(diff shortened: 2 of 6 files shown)


angeloashmore:
This looks like a good change! As @dani-mp said, we can treat this as a bug fix without waiting for the content relationship updates. `isFilled.contentRelationship` should support unions of `ContentRelationshipField`.

I suggest we make the following changes:

- Don't overload `contentRelationship`. Instead, only include the simplified version introduced in this PR:
   ```ts
   export const contentRelationship = link as <
   	Field extends ContentRelationshipField,
   >(
   	field: Field | null | undefined,
   ) => field is Extract<Field, FilledContentRelationshipField>
   ```

   We don't document the type parameters, and I haven't seen anyone use them in the wild. They are an implementation detail. Thus, I don't think we need to support the old type parameters.

- Modify `FilledContentRelationshipField` to make the `data` property optional only when `DataInterface` is not populated.

   ```ts
   export type FilledContentRelationshipField<
   	TypeEnum = string,
   	LangEnum = string,
   	DataInterface extends
   		| Record<string, AnyRegularField | GroupField | SliceZone>
   		| unknown = unknown,
   > = {
   	link_type: "Document"
   	id: string
   	uid?: string
   	type: TypeEnum
   	tags: string[]
   	lang: LangEnum
   	url?: string
   	slug?: string
   	isBroken?: boolean
   	// eslint-disable-next-line @typescript-eslint/no-explicit-any
   } & (DataInterface extends Record<string, any>
   	? { data: DataInterface }
   	: { data?: unknown })
   ```

   This allows for using `document.fooOrBar.data` when `DataInterface` is provided.

With these changes, we should be able to generate types like your example `fooOrBar` property. How we generate the `DataInterface` types doesn't matter (e.g. via standalone types or complex `Pick`).

angeloashmore (changes requested):
This looks like a good change! As @dani-mp said, we can treat this as a bug fix without waiting for the content relationship updates. `isFilled.contentRelationship` should support unions of `ContentRelationshipField`.

I suggest we make the following changes:

- Don't overload `contentRelationship`. Instead, only include the simplified version introduced in this PR:
   ```ts
   export const contentRelationship = link as <
   	Field extends ContentRelationshipField,
   >(
   	field: Field | null | undefined,
   ) => field is Extract<Field, FilledContentRelationshipField>
   ```

   We don't document the type parameters, and I haven't seen anyone use them in the wild. They are an implementation detail. Thus, I don't think we need to support the old type parameters.

- Modify `FilledContentRelationshipField` to make the `data` property optional only when `DataInterface` is not populated.

   ```ts
   export type FilledContentRelationshipField<
   	TypeEnum = string,
   	LangEnum = string,
   	DataInterface extends
   		| Record<string, AnyRegularField | GroupField | SliceZone>
   		| unknown = unknown,
   > = {
   	link_type: "Document"
   	id: string
   	uid?: string
   	type: TypeEnum
   	tags: string[]
   	lang: LangEnum
   	url?: string
   	slug?: string
   	isBroken?: boolean
   	// eslint-disable-next-line @typescript-eslint/no-explicit-any
   } & (DataInterface extends Record<string, any>
   	? { data: DataInterface }
   	: { data?: unknown })
   ```

   This allows for using `document.fooOrBar.data` when `DataInterface` is provided.

With these changes, we should be able to generate types like your example `fooOrBar` property. How we generate the `DataInterface` types doesn't matter (e.g. via standalone types or complex `Pick`).

src/helpers/documentToLinkField.ts
```diff
@@ -45,5 +45,9 @@ export const documentToLinkField = <
 		...(prismicDocument.data && Object.keys(prismicDocument.data).length > 0
 			? { data: prismicDocument.data }
 			: {}),
-	}
+	} as FilledContentRelationshipField<
+		TDocument["type"],
+		TDocument["lang"],
+		TDocument["data"]
+	>
```
lihbr, line 52: I'm not satisfied by that, but the error is cryptic and I don't understand what's throwing TypeScript off:

```
Type '{ data?: Record<string, any> | undefined; link_type: "Document"; id: string; uid: string | undefined; type: string; tags: string[]; lang: string; url: string | undefined; slug: string | undefined; }' is not assignable to type 'FilledContentRelationshipField<TDocument["type"], TDocument["lang"], TDocument["data"]>'.
  Type '{ data?: Record<string, any> | undefined; link_type: "Document"; id: string; uid: string | undefined; type: string; tags: string[]; lang: string; url: string | undefined; slug: string | undefined; }' is not assignable to type 'TDocument["data"] extends Record<string, any> ? { data: TDocument["data"]; } : { data?: unknown; }'.ts(2322)
```

angeloashmore: We can't differentiate between a provided data type (e.g. `{ foo: NumberField }`) and the default `DataInterface` (i.e. `Record<string, AnyRegularField | GroupField | SliceZone>`). Thus, I don't think we can write a correct implementation or overload without `as`.

Is there any case where `data` might not be filled at runtime? Our code contains the following comment:

```ts
// The REST API does not include a `data` property if the data
// object is empty.
```

- If **yes**: Let's keep the existing `data?:` property and remove the `as`. Consumers will need to write `document.contentRelationship.data?.foo`, which is correct since `data` may not be included.
- If **no**: Let's use the type assertion (`as`). We know the output is correct, and we have tests to verify. Consumers can write `document.contentRelationship.data.foo`.

lihbr: Turns out it's still a possible behavior, so let's revert: https://200629-sms-hoy.cdn.prismic.io/api/v2/documents/search?q=%5B%5Bat%28document.type%2C+%22kitchen_sink_4%22%29%5D%5D&ref=aDbsjxEAACwAkpaF&fetchLinks=post.title

PR #396, opened 2025-08-14
add x-client headers to access migration api

dani-mp:
Resolves: 
https://linear.app/prismic/issue/DT-2838/migration-api-client-add-a-header-to-inform-us-prismic-who-is-using

### Description

we want to track if the migration api is used via our sdk or directly - so we're adding a couple of headers when making requests to the migration api

### Checklist

### How to QA [^1]

[^1]:

### DIFF
--- a/src/WriteClient.ts
+++ b/src/WriteClient.ts
@@ -32,6 +32,8 @@ import { InvalidDataError } from "./errors/InvalidDataError"
 import { NotFoundError } from "./errors/NotFoundError"
 import { PrismicError } from "./errors/PrismicError"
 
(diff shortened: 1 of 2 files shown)


src/WriteClient.ts
```diff
@@ -769,6 +771,8 @@ export class WriteClient<
 			}),
 			headers: {
 				"content-type": "application/json",
+				"x-client": "@prismicio/client",
```
lihbr, line 774: 💡 #idea: For consistency, I'd maybe favor importing `name` from the package here also to prevent future typos

Issue #399, opened 2025-09-25
Support type checking on nodenext


cardosod-ppb:
### Is your feature request related to a problem? Please describe.

I'm unable to validate types using module resolution `nodenext`, since it expects the inclusion of the file extension on the import statements.

Consider the following example:

```ts
import {  BooleanField } from "@prismicio/client";

type Foo = {
  bar: BooleanField;
}

const foo: Foo = {
  bar: "true", // This should throw an error
}
```

### Describe the solution you'd like

I would suggest including the file extension on all the import statements.

The previous example works fine with the following change:

```ts
// src/index.ts
export type { BooleanField } from "./types/value/boolean.ts"
```

PR #405, opened 2025-10-15
build: migrate to `unbuild` to support modern requirements

lihbr:
Resolves: #399

### Description

This PR experiments with the following to fix modern TypeScript build requirements:
- Migrate to `unbuild`
- Review exports based on `ofetch`: https://github.com/unjs/ofetch/blob/main/package.json
- Update `vitest` to the latest version

This changes the structure of the `dist` folder quite a lot, so we'd have to test if it's compatible with various technologies, especially React and Next.js

### Checklist

### Preview

N/A

### How to QA [^1]

`@prismicio/client@alpha`

[^1]:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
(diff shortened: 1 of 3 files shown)


angeloashmore:
Changing build tools is a risky step as it could have unintended side effects. unbuild is also positioned to be replaced by obuild, or more realistically, by Vite/tsdown.

To solve #399 specifically, I think we can make a less invasive change, like automatically adding `.js` to imports and exports in generated `.d.ts` files.

See #406 and https://github.com/prismicio-community/vite-plugin-sdk/pull/11

PR #442, opened 2026-04-02
fix: Use arrayBuffer instead of blob when memoizing responses

anakinj:
Resolves: React native compatibility 

### Description

The react native blob implementation is apparently a bit lacking. The memoization introduced in #437 is using some functions that do not exist in the react native blob.

Resulting in the following errors:
```
TypeError: _blob.text is not a function (it is undefined)
```

This change uses the `response.arrayBuffer()` combined with `TextEncoder`.

### Checklist

### How to QA [^1]

[^1]:

---

> [!NOTE]
> **Medium Risk**
> Changes the core `request()` deduplication/memoization path to buffer bodies as `ArrayBuffer`, which could affect response decoding and memory usage across all callers. Risk is moderate because it touches shared HTTP plumbing but is a localized behavioral swap.
> 
> **Overview**
> Fixes response memoization to be React Native compatible by buffering deduplicated responses using `response.arrayBuffer()` instead of `response.blob()`.
> 
> `memoizeResponse()` now reconstructs `text()`/`json()` via `TextDecoder`, exposes `arrayBuffer()`, and recreates `blob()` from the buffered bytes; `WriteClient.fetchForeignAsset()` is updated to build `File` objects from an `ArrayBuffer` rather than a `Blob`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 71c6c7eefea13e189fc275dea71b469672fe434f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/WriteClient.ts
+++ b/src/WriteClient.ts
@@ -633,10 +633,10 @@ export class WriteClient<
 			throw new PrismicError("Could not fetch foreign asset", url, undefined)
 		}
 
-		const blob = await res.blob()
+		const buffer = await res.arrayBuffer()
 
 		// Ensure a correct content type is attached to the blob.
-		return new File([blob], "", {
+		return new File([buffer], "", {
 			type: res.headers.get("content-type") || undefined,
 		})
 	}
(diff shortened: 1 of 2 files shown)


lihbr (approved):
LGTM, thank you so much for taking the time to open a PR! 💙

angeloashmore (approved):
Looks good to me too! Thanks for the fix. 🙂

PR #446, opened 2026-08-04
feat: add WriteClient.publishMigrationRelease()

tomaboro:
Resolves: [DT-3509](https://linear.app/prismic/issue/DT-3509/publish-a-migration-release-via-the-migration-api)

## Summary

Adds a public `publishMigrationRelease()` method to the write-capable `WriteClient`, letting users publish their repository's migration release programmatically.

- POSTs to `migration-release/publish` on the Migration API endpoint via the shared private request helper 
- Takes no arguments (optional fetch params only) and returns `{ totalItems: number }` — the number of documents queued for publishing.
- Publishing is asynchronous on the backend, so content goes live shortly after the call returns.

Made with [Cursor](https://cursor.com)

BEGIN_COMMIT_OVERRIDE
feat: add `WriteClient.publishMigrationRelease()`
END_COMMIT_OVERRIDE


test/createWriteClient.test.ts
```diff
@@ -44,3 +44,11 @@ it("supports custom migration api endpoint", async ({ expect, writeToken }) => {
...
 	expect(res.migrationAPIEndpoint).toBe("https://example.com/")
 })
+
+it("exposes a publishMigrationRelease method", async ({
+	expect,
+	writeToken,
+}) => {
+	const res = createWriteClient("example", { writeToken })
+	expect(typeof res.publishMigrationRelease).toBe("function")
+})
```
angeloashmore: ⚠️ #issue: This test isn't necessary. The `test/writeClient-publishRelease.test.ts` implicitly checks presence.

test/writeClient-publishRelease.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+import { version } from "../package.json"
+import { ForbiddenError, NotFoundError, PrismicError } from "../src"
+import { it } from "./it"
+
+describe("publishMigrationRelease", () => {
+	it("publishes the migration release and returns the total", async ({ expect, writeClient }) => {
+		vi.mocked(writeClient.fetchFn).mockResolvedValueOnce(
+			Response.json({ totalItems: 3 }, { status: 202 }),
+		)
```
angeloashmore: ⚠️ #issue: The client's tests use a real repository and network requests. We avoid mocking responses, even if it speeds up tests. In addition to making sure we're calling the right endpoint, we also want to make sure the client works with the real deployed API.

We typically will mock requests only for different status codes since we can't always replicate a scenario.

Since the migration release is a repository singleton, we can either:
1. Create a new repository in each test. See https://github.com/prismicio/cli/blob/74454bdebdc7f098235f0d5b461eabda38802160/test/it.ts#L171-L182 for an example implementation.
2. Run tests serially, ensuring each test doesn't interfere with others.

We can either check the API response (number of published items) or verify that the Content API returns the published items. Your call depending on how E2E-y you want to get.

test/writeClient-publishRelease.test.ts
```diff
@@ -0,0 +1,110 @@
...
+	}) => {
+		vi.mocked(writeClient.fetchFn).mockResolvedValueOnce(
+			Response.json({ totalItems: 0 }, { status: 202 }),
+		)
+
+		await writeClient.publishMigrationRelease()
+
+		const init = vi.mocked(writeClient.fetchFn).mock.calls[0][1]
+		expect(init?.body).toBeTruthy()
+	})
```
angeloashmore: ❓ #ask: Testing implementation details like this seems like a code smell. WDYT?

angeloashmore (commented):
Implementation looks clean! I have comments about how we are testing the new `publishMigrationRelease` method.

src/WriteClient.ts
```diff
@@ -235,6 +236,54 @@ export class WriteClient<
...
+	): Promise<PublishMigrationReleaseResult> {
+		const url = new URL("migration-release/publish", this.migrationAPIEndpoint)
+
+		const response = await this.#request(url, params, {
+			method: "POST",
+			headers: {
+				"content-type": "application/json",
+				"x-client": CLIENT_IDENTIFIER,
+			},
+		})
```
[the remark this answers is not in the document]
tomaboro: Fixed in 5cc97b1 — the POST now sends a JSON body so `request()` treats it as a write, routing it through the per-host throttle instead of the read-dedup path.

test/createWriteClient.test.ts
tomaboro: Agreed, removed — presence is implicitly covered once the publish tests call the method.

test/writeClient-publishRelease.test.ts
tomaboro: Switched to option 2: publish tests now run against an isolated repository created via `repositories.createRepository()`, using the real Migration API for the happy path (create + migrate + publish, asserting `totalItems`). Kept mocks only for the 404/500 error cases we can't otherwise reproduce.

test/writeClient-publishRelease.test.ts
tomaboro: Agreed, removed that test.

angeloashmore (approved):
Approved, but I opened a PR on top of this one that makes the following organizational edits: https://github.com/prismicio/prismic-client/pull/447

- Renames `test/writeClient-publishRelease.test.ts` to `test/writeClient-publishMigrationRelease.test.ts`.
- Refactors `isolatedWriteClient` to a fixture, reusable anywhere.

Could you take a look at it and, if it looks good, merge into this PR before merging? Not blocking if you decide against it. 🙂

